PR #286, opened 2018-12-04
Feature/implement history filter

### LINKED ISSUE #233: Add filter to History tool
### Description
The History tool displays http requests that have been sent since the target page was was loaded. Often there can be a lot of messages, and you may only want to see certain types of messages. To help this we want to add support to add a search or regex feature that would remove messages that didn't match.

There are a number of different ways we could support this, but I think the most cohesive way is to follow the way dev tools work with a single search bar that supports searching and filtering:
image

### Requirements
* This change will likely be limited to tool/history.js, drawer.js, drawer.html, drawer.css.
* Need to design the search bar element.
* Need to be able to persist the filter between page loads (use localstorage)
* Should perform _some_ filtering, searching, or fuzzy matching to give the expected results.

jaywon:
Main functionality to resolve #233 :

* Sticky header for field titles and filter
* Support for both in string filter matches and regular expressions, including visual invalid RegEx state styling and toggle
* Styling for smaller font-size and table element spacing
* Filtered/total message count awareness
* Bug fix for document.getElementById error when no messages are visible with filter when targeting last message element as new messages load


psiinon:
Thanks for this :) Will review as soon as I get a chance...

src/main/zapHomeFiles/hud/drawer.html
```diff
@@ -20,24 +21,50 @@
...
+                            </span>
+                            <span>
+                                <input v-model="regexEnabled" type="checkbox" />
+                            </span>
+                            <span>
+                                Enable RegEx
+                            </span>
+                        </th>
+                        <th class="title hidden">
+                            <span>{{ hiddenMessageCount }} of {{ messageCount }} items hidden by filters</span>
```
psiinon: Should be i18n'd

src/main/zapHomeFiles/hud/drawer.html
```diff
@@ -20,24 +21,50 @@
...
+                        </th>
+                        <th class="title filter"> 
+                            <span>
+                                <input v-model="filter" type="text" placeholder="Filter" />
+                            </span>
+                            <span>
+                                <input v-model="regexEnabled" type="checkbox" />
+                            </span>
+                            <span>
+                                Enable RegEx
```
psiinon: Should be i18n'd

jaywon:
Let me know. I had pushed and opened PR to the private repo before the multi-tab support got merged to `develop` and repo switched. After that got merged and merging into my fork i wasn't seeing history messages being dispatched at all, and haven't confirmed yet if it was an issue w/ my environment or something about the merge. 

Will try to get some time into it this weekend to troubleshoot further. Also want to start looking at tests around the API for verifying these kinds of things.

jaywon:
@psiinon Yup! Will make those changes and I did want to reduce the font size and add a draggable bar to let the user manage visibility based on personal preference. 👍

psiinon:
We should probably have a 'global' control to manage font size, as this is something that should apply to all tools.

src/main/zapHomeFiles/hud/drawer.js
```diff
@@ -12,19 +12,57 @@ Vue.component('history', {
...
+            const self=this,
+                  isRegex = this.regexEnabled;
+            let re;
+
+            if (isRegex){
+                try {
+                    re = new RegExp(this.filter);
+                }
+                catch (ex) {
+                    re = new RegExp(''); //hack for now to prevent errors being thrown for invalid RegExp
```
dscrobonia: what would be the "non hack" solution? would it be to also alert the user of invalid regex?

psiinon:
I quite like the hidden messages counter ;)

jaywon:
Gonna make those changes today. I actually took the hidden messages concept from Firefox Console tab filter 😄

src/main/zapHomeFiles/hud/drawer.js
jaywon: I was thinking the "non hack" solution would be to add some visual indicator to the user that the current RegEx was invalid, like a red outline around the input box? Would be easy to wire up to a state in this `catch` as they type out and can easily understand why no messages are showing. Kind of thinking how wireshark changes from green background to red as your typing in filters but more subtle. What you think?

I could do it on this PR, was really trying to get this in before launch but since I missed that I don't mind taking a little more time for polish.

jaywon:
@thc202 Sure! Can you clarify where you'd like to see that?  The description on this PR or is there somewhere else you'd like this documented?

thc202:
In [CHANGELOG.md](https://github.com/zaproxy/zap-hud/blob/develop/CHANGELOG.md), worth rebase/merge the latest changes to avoid conflicts.

dscrobonia:
Do we have notes somewhere about how to use the changelog? this is new to me.

src/main/zapHomeFiles/hud/drawer.js
dscrobonia: I think that's a great idea! if you can add that in without too much more effort I think that'd be a great addition. :)

psiinon:
btw I'm raised my previous comment about font sizes as a separate issue #355

jaywon:
@psiinon Lemme look into that bug. I made a few more small UI/UX tweaks and I think w/ that bug squashed this should address all feedback and hopefully in a good merge state!

jaywon:
@psiinon @dscrobonia @thc202 Think I addressed all feedback from the team including addition to CHANGELOG, styling issues, latent count display, and i18n support. Still tweaks to be made but I'm hoping that this is in a good enough state to merge and create new PR(s) for additional tweaks.

src/main/zapHomeFiles/hud/drawer.html
```diff
@@ -20,24 +21,50 @@
...
+                            </span>
+                            <span>
+                                <input v-model="regexEnabled" type="checkbox" />
+                            </span>
+                            <span>
+                                {{ enableRegExText }}
+                            </span>
+                        </th>
+                        <th class="title hidden">
+                            <span>{{ hiddenMessageCount }} {{ commonOf }} {{ messageCount }} {{ historyItemsFilteredSuffix }}</span>
```
psiinon: Just noticed this - we're assuming the text structure is "x of y etc" which is obviously right in English but might not be right in other languages, eg Japanese / Chinese etc? 
I think the solution is to have a parameterized i18n string like: 
history_of = {0} of {1} items hidden by filters
and to pass the values as params

jaywon: Good catch! I'll take care of that today 👍

jaywon: @psiinon Updated w/ latest request!

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+ - Add plain text/regex filtering capability to History section [#286](https://github.com/zaproxy/zap-hud/issues/233)
```
thc202: This should be added under the `Added` section below.

src/main/zapHomeFiles/hud/drawer.js
```diff
@@ -12,19 +12,73 @@ Vue.component('history', {
...
+                    return re.test(message.url);
+                }
+                else{
+                    return message.url.indexOf(self.filter)>=0; 
+                }
+            });
+        },
+        historyItemsFiltered() {
+            let historyItemsFilteredString = I18n.t("history_items_filtered")
+                    .replace('{0}', this.hiddenMessageCount)
```
thc202: With latest changes (merge upstream) these can be passed as parameters to `t`, like in:
https://kazupon.github.io/vue-i18n/guide/formatting.html#list-formatting

src/main/zapHomeFiles/hud/drawer.js
```diff
@@ -12,19 +12,73 @@ Vue.component('history', {
...
     },
+    watch: {
+        filter() {
+            //use the next tick capability to let the filteredMessages state update displayed messages and update count after reactive update
+            this.$nextTick(function () {
+                const visibleMessages = document.querySelectorAll("#history-messages .message-tr");
+                this.hiddenMessageCount = (!this.messages) ? 0 : this.messages.length - visibleMessages.length;
+            })
+        },
+        regexEnabled(){
```
thc202: Changing just the regex state does not affect the message that shows the number of items filtered, it keeps showing the old state even though the messages are filtered.

src/other/resources/UIMessages/UIMessages.properties
```diff
@@ -59,6 +59,7 @@ common_start = Start
 common_stop = Stop
 common_turn_on = Turn On
 common_turn_off = Turn Off
+common_of = of
```
thc202: Remove? Does not seem to be in use anymore.

src/main/zapHomeFiles/hud/drawer.js
jaywon: Validated that works! I had tried that method previously but the interpolation wasn't working before. Merging upstream fixed, thanks for heads up, felt wrong. 👍

src/main/zapHomeFiles/hud/drawer.js
jaywon: Implemented in latest revisions!

src/main/zapHomeFiles/hud/drawer.js
jaywon: @thc202 nice catch! I refactored for the reactivity to trigger on the filtered message list rather than the filter input itself which made for a better solution of handling the various use cases.

thc202:
Thanks!

PR #287, opened 2018-12-04
Disable CSS format check for the time being

thc202:
Disable to not require npm when trying the HUD directly from the repo.


psiinon:
Thanks!

PR #303, opened 2018-12-17
Add option to control on-domain messages

### LINKED ISSUE #294: Add option to not run tools on target domain
Eg the input field highlighting and reveal tool.
If turned on we would do the minimum changes to the target domain possible and not accept any operations from it.
This would add extra protection from potentially malicious sites.

psiinon:
Fixes  #294


psiinon:
Arg, I'll get used to this new dev route at some point...

psiinon:
Added changelog entry

Issue #305, opened 2018-12-17
Package the client side libraries


psiinon:
Eg using npm and packages.json

psiinon:
Libs that we dont appear to be tracking (via https://github.com/zaproxy/zap-hud/network/dependencies) include:

- alertify.js (which isnt in NPM and we need to replace: #306)
- localforage
- vue-i18n

PR #307, opened 2018-12-18
Added localforage and vue-i18n

psiinon:
Related to #305 - the idea is that just adding these libs will mean we get github notifications if any security issues are found with them.
I'm new to NPM etc, so please check this carefully :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,9 @@
   },
   "homepage": "https://github.com/zaproxy/zap-hud/wiki",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


jaywon:
Just as an FYI based on the comment of checking carefully. This is correct that Github scans the `package.json` dependencies and does their audits with this tool that got pulled into NPM official earlier this year.

https://github.com/nodesecurity/nsp

It's a command line tool that we could build into any CI/auditing mechanisms as well if needed.

PR #310, opened 2018-12-20
Add task to install the add-on into running ZAP

thc202:
Add a task `installAddOn` that installs the add-on into ZAP (started
with `runZap`), to allow to apply add-on changes without restarting ZAP
(which takes more time than un/install the add-on).

---
Requires zaproxy/zaproxy#5157.


thc202:
Rebased to trigger the latest checks.

PR #311, opened 2018-12-21
Remove stale add-ons when building

thc202:
Change `assembleZapAddOn` to remove stale add-ons, to ensure ZAP loads
the add-on built.

### DIFF
--- a/buildSrc/src/main/java/org/zaproxy/gradle/AddOnPlugin.java
+++ b/buildSrc/src/main/java/org/zaproxy/gradle/AddOnPlugin.java
@@ -19,7 +19,13 @@
  */
 package org.zaproxy.gradle;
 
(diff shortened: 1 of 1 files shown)


kingthorin:
Build failed?

thc202:
Yes, but it's unrelated to these changes, the tutorial tests fail sometimes (restarted the failing job).

thc202:
All passing now.

dscrobonia:
I haven't looked too much into the selenium tests, but is there a reason why they fail sometimes? Is there a way we can make those more reliable? I know @psiinon did a ton of work getting those up, so I think it would be cool if we could make those trustworthy. Right now they don't really provide me value because if they fail I can't trust taht something is actually wrong.

PR #313, opened 2018-12-26
Wait for the panel to be stale in tutorial tests

thc202:
Change `HUD.openUrlWaitForHud` to wait for the left panel to become
stale when the HUD is being initialised (service worker installed and
page refreshed) and wait for the page to complete loading, as it is more
reliable than checking the number of buttons loaded in the panel.


dscrobonia:
Late to the party, but :clap: :smile: :tada:  for more reliable tests! You're the best @thc202 !

PR #314, opened 2018-12-27
Require cookie on all API calls

psiinon:
More defence in depth - this will prevent malicious sites from directly accessing the ZAP API, even if they somehow manage to get hold of the API URL (which they shouldn't be able to).


psiinon:
We still have the random path element - this is in addition.
btw the tests seem to be failing consistently on JDK 11 - I need to look into that...

psiinon:
Fixed conflicts and hopefully the failing test (thanks @thc202)

psiinon:
Ugh, missed out some important changes :/ lets see if this is any better..

src/main/java/org/zaproxy/zap/extension/hud/HudApiProxy.java
```diff
@@ -64,11 +68,26 @@ public void addCustomHeaders(String name, RequestType type, HttpMessage msg) {
...
+        String url = msg.getRequestHeader().getURI().toString();
         try {
 
-            String url = msg.getRequestHeader().getURI().toString();
             LOG.debug("API proxy callback url = " + url);
 
+            if (!api.getZapHudCookieValue()
+                    .equals(api.getRequestCookieValue(msg, HudAPI.ZAP_HUD_COOKIE))) {
+                String errorMsg =
+                        "Correct cookie not supplied when accessing API "
```
kingthorin: The negative logic in this error message is difficult. Perhaps "Incorrect cookie supplied when accessing the API "

src/main/java/org/zaproxy/zap/extension/hud/HudApiProxy.java
```diff
@@ -64,11 +68,26 @@ public void addCustomHeaders(String name, RequestType type, HttpMessage msg) {
...
+                                + msg.getRequestHeader().getURI();
+                if (url.indexOf("hud/action/log") > 0) {
+                    // This will happen on start up, especially in dev mode, so just fail quietly
+                    LOG.debug(errorMsg);
+                    return "";
+                }
+                LOG.warn(errorMsg);
+                throw new ApiException(
+                        ApiException.Type.ILLEGAL_PARAMETER, "Correct cookie not supplied");
```
kingthorin: Re-use `errorMsg`?

src/main/java/org/zaproxy/zap/extension/hud/HudApiProxy.java
psiinon: +1

PR #320, opened 2018-12-28
Add API endpoints for getting and setting UI options

### LINKED ISSUE #319: Add API endpoints for getting and setting UI options
These should cope with arbitrary keys so that the UI can get/set whatever options it needs to without any further Java changes.
The UI options are currently persisted in the browser session, but this means they are not persisted when 'browser launch' is used.

psiinon:
Fixes #319 

@dscrobonia are these endpoints suitable for getting and setting the UI options?
The values are just strings, but they could be parsed as json / whatever if thats useful.
Theres no upper limit to the size of the values that can be set, but I'd suggest keeping them relatively small if possible.


psiinon:
@dscrobonia can always change things once you have had a chance to use them :)

PR #327, opened 2018-12-31
Add 2nd non SameSite cookie for safer API calls - WIP

### LINKED ISSUE #326: Tutorial 'Dont show this screen again' option broken
Looks like this is because the new SameSite cookie is not being added to the API request.
Apparently this is because the ZAP frame is on the target site, but we have other similar cases (eg the Site tree) where the API calls appear to still work.
Am investigating...

psiinon:
Fixes #326 

The reason for the "WIP" is because it looks to me like some of the API requests coming from the display do have the SameSite cookie attached :/
A good example is the Sites tree tool - the display appears to be making direct API calls https://github.com/zaproxy/zap-hud/blob/develop/src/main/zapHomeFiles/hud/display.js#L458 and these do have the SameSite cookie :/

Thoughts appreciated...

### DIFF
--- a/src/main/java/org/zaproxy/zap/extension/hud/HudAPI.java
+++ b/src/main/java/org/zaproxy/zap/extension/hud/HudAPI.java
@@ -63,7 +63,8 @@ import org.zaproxy.zap.extension.websocket.ExtensionWebSocket;
 
 public class HudAPI extends ApiImplementor {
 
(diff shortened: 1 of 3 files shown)


psiinon:
Will also need to update the changelog of course...

psiinon:
Closing this as it wont be needed once #358 has been merged.

PR #341, opened 2019-01-04
Use websockets for all API calls

### LINKED ISSUE #129: Scanner tools shown as started even if not
The scanner tools (e.g. spider, active scanner) are shown as started (e.g. `0%` progress or with coloured icon) even if they were not started successfully (because of an error). This can be reproduced easily by setting the mode to Safe and then attempt to start a scan.

psiinon:
Requires https://github.com/zaproxy/zap-extensions/pull/1906 so wont work unless you have rebuilt the websocket add-on with that PR.

The plan is to replace all HTTPS API calls with websocket ones.
Will also check all of the responses as part of this change.

Fixes #129


psiinon:
Have changed apiCall to return a promise and fixed the conflicts.
@dscrobonia that look ok to you?
I still plan to change the apiCall i/d to use parameters rather than an object, and need to convert the remaining calls, but hopefully the general principal looks good.

psiinon:
Updated to use arrow functions and fail the promse on an error response. Have also split the apiCall into 2 functions, one now called apiCallWithResponse.
This look ok before I apply the changes to the remaining https API calls?

psiinon:
This pull request **introduces 1 alert** when merging 2b58a2db53a591779a085cdd167d42598c971dfa into d2115cee8dae0ec5da9cd028da7eb2e24f3d8e06 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-hud/rev/pr-7b27421024ac0c5b43ad4cb6dbf3357645d86110)

**new alerts:**

* 1 for Unused variable, import, function or class

---

src/main/zapHomeFiles/hud/display.js
```diff
@@ -386,16 +386,18 @@ Vue.component('history-message-modal', {
 		replayInBrowser: function() {
 			let self = this;
 			let message = this.request;
-			utils.zapApiCall("/hud/action/recordRequest/?header=" + encodeURIComponent(message.header) + "&body=" + encodeURIComponent(message.body))
-			.then(response => response.json())
-			.then(json => {
-				if (json.requestUrl) {
-					window.top.location.href = json.requestUrl;
+			let channel = new MessageChannel();
```
dscrobonia: do you think we should have a function to wrap up this channel logic? idk may be too much in other frames.

src/main/zapHomeFiles/hud/serviceworker.js
```diff
@@ -136,10 +138,24 @@ const onMessage = event => {
 			break;
 
 		case "heartbeat":
-			webSocket.send('{ "component" : "hud", "type" : "view", "name" : "heartbeat" }');
+			apiCall("hud", "view", "heartbeat");
 			break;
 
+		case "zapApiCall":
+			if (event.ports.length > 0) {
+				apiCallWithResponse(message.component, message.type, message.name, message.params).then (response => {
```
dscrobonia: the style throughout the rest of the code is to have the `.then(response => {` start on the next line. helps to line up the promises. 

eventually all promises should be handled with:
```
let response = await apiCallWithResponse(...);
event.ports[0].postMessage(response);
```
which is obviously way nicer.

src/main/zapHomeFiles/hud/serviceworker.js
```diff
@@ -175,7 +203,33 @@ webSocket.onerror = function (event) {
...
+	let call = { component : component, type: type, name: name, params: params }; 
+	webSocket.send(JSON.stringify(call));
+};
+
+function apiCallWithResponse(component, type, name, params) {
+	if (! params) {
+		params = {};
+	}
+	let call = { component : component, type: type, name: name, params: params }; 
+	var pFunctions = {};
```
dscrobonia: let > var

src/main/zapHomeFiles/hud/serviceworker.js
```diff
@@ -167,6 +183,18 @@ webSocket.onmessage = function (event) {
...
 		var ev = new CustomEvent(jevent['event.publisher'], {detail: jevent});
 		self.dispatchEvent(ev);
+	} else if ('id' in jevent && 'response' in jevent) {
+		let pFunctions = webSocketCallbacks[jevent['id']];
+		let response = jevent['response'];
+		if ('code' in response && 'message' in response) {
+			// These always indicate a failure
+			pFunctions.reject(response);
+		} else {
+			pFunctions.resolve(response);
```
dscrobonia: maybe in the failure case we should wrap the response in an `Error` object and supply the response as event data?

src/main/zapHomeFiles/hud/utils.js
```diff
@@ -497,7 +497,16 @@ var utils = (function() {
 			})
 			.catch(errorHandler);
 	}
-	
+
+	function zapApiErrorDialog(tabId,  response) {
```
dscrobonia: is there a place where we'll use the error dialog vs just logging the error? or should we put it all failure cases? curious if you had thought about that or if you justed wanted to surface these error messages.

src/main/zapHomeFiles/hud/serviceworker.js
psiinon: will do

src/main/zapHomeFiles/hud/serviceworker.js
psiinon: will realign

src/main/zapHomeFiles/hud/serviceworker.js
psiinon: yeah, looks like wrapping the response in an Error is the 'standard' way to go.

src/main/zapHomeFiles/hud/utils.js
psiinon: I think it all depends on the caller - this is a log level function so I dont think it should dictate how the failure is handled

psiinon:
Just moved the active scan calls to use websockets.
I plan to move all of the api calls, but if the tests pass I'll remove the WIP bit so this can be merged. I'll then open new PRs for moving the other calls.

thc202:
The Java 11 failure is being addressed in #356.

psiinon:
Have removed the WIP - I think its ready for review and can be merged once it passes those and #356 is merged.

src/main/zapHomeFiles/hud/display.js
dscrobonia: Agreed.

PR #350, opened 2019-01-09
Fix running tab bug

dscrobonia:
because the active scan and spider tool changed button state on tabId, and not by domain, when a user visits a new page within the same site the corresponding button would be in the wrong state and unclickable. 

Now active scan and spider tool show the state of the buttons and update progress by scope. should also update support to scan via the scope, and not just the current domain its on. Attack mode also works a little differently and they could be made consistent.


psiinon:
This pull request **introduces 1 alert** when merging e79c24f55bb101711b1492fcbf178068b637add1 into d2115cee8dae0ec5da9cd028da7eb2e24f3d8e06 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-hud/rev/pr-75d986194700f9886ef3ac779630dd1043b8f319)

**new alerts:**

* 1 for Unneeded defensive code

---

psiinon:
Looks good :)
There is one issue (apart from the LGTM one):
If you start an active scan on one tab then the same control in another tab on the same domain allows you to stop it, which is good.
However if you start the spider then the spider controls on the other tabs on the same domain are disabled. I think these should stay enabled as with the active scan - this allows you to stop the spider if you close the tab you started it on.

psiinon:
Hum, another thought - do we need to disable the control on tabs on other domains?
Why not show the domain being spidered / ascanned (in the dialog) and give the option to stop it?

dscrobonia:
Yeah you make a good point, why not just allow these tools be turned on or off from any domain? ... The only reason we would want to keep them running on domain would be to clarify which domains the tool is running on I think. Because theoretically you could start a scan on one domain, close that tab, and not be able to stop the scan if you weren't on that in scope page..... which doesn't really make sense. 

Yeah maybe we should be able to stop on all domain. But I would want some sort of indicator to know what were scanning while it happens. Like maybe we should have some sort of global "status" indicator. So, for example, when we're active scannig the full scope, in the upper left hand corner we have a growler-esque notification that lets us know were active scanning a specific scope. While we're scanning it remains there, but when we're done it goes away. Or when we active scan a specific url, the global scope notification show us that. Would also work for spidering, attack, fuzzing, etc... Basically any background activity would keep its status updated in this global status indicator while we do other  things. 

I know the tools are supposed to let the user know of their status via their data field. but idk if thats always enough. Or maybe we should still use the tool's "data field" but add more data (like instaed of 40% scanned in the active scan tool, we would say "example.com 50% scanned". Like after starting the spider i want to know what the scope of what I was spidering was and the progress of it is. (it won't always be just whats in scope via the scope tool, we might want to spider a specific url) so I want to know whats happening in the backgroudn while im visiting sites.

Anyways, lmk you're thoughts.

psiinon:
So why not put this info in the dialogs for the tools?
So if you click the Spider tool while its running you get a dialog with:

> The spider is currently running:
> Target: https://example.com
> URLs found: 2102
> Would you like to stop it?
> [Stop] [Cancel]

And once its stopped and you click on it again then you'll see:

> Start spidering https://new-target.com?
> Previous target: https://example.com
> URLs found: 7392
> [Start] [Cancel]

dscrobonia:
@psiinon lmk how this looks for ya. Should be good to go now! :)

I've left off the past scan results for now. I think including all of that in the text dialog might be a little overwhelming. I do think we need to figure out the best way to incorporate that, but lets do it in another PR.

psiinon:
This pull request **introduces 1 alert** when merging 509c500ac14db9a426618dffad889a5c0ade4f51 into ded3bb7ea76aac0aad33e6b68d05711aaa41fe71 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-hud/rev/pr-e071d4f0e4561b6c482e15dcbe5803010467e58a)

**new alerts:**

* 1 for Unneeded defensive code

---

psiinon:
See above for new alerts c/o LGTM ;)
Also when stopping the Spider get:

> The site undefined is currently getting Spidered.
> Would you like to stop the Spider tool?

Don't think it should be 'undefined' ;)
The active scan one works fine.

dscrobonia:
The LGTM alert is a false-positive. did confirm. lmk how I should deal with that. :)

psiinon:
This pull request **introduces 2 alerts** when merging 4164bb28310f81aaded64d1b733fd2cb0007dcd9 into c8bf09d24cc87b62fd2b4a865ef43a44430baa19 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-hud/rev/pr-d5190a4e0acda6029490810b2b7c639f06e76b43)

**new alerts:**

* 1 for Superfluous trailing arguments
* 1 for Unneeded defensive code

---

psiinon:
I'm not sure it is a false positive - it looks like there is no function overloading in javascript: https://stackoverflow.com/questions/456177/function-overloading-in-javascript-best-practices
I think the best option is to either have one function addToScope(tabId, domain) and test to see if domain is not supplied or to have 2 different function names.

dscrobonia:
@psiinon I think thi sis good now

psiinon:
It still has LGTM errors.
And when I start the active scanner the status isnt getting updated - its staying as 'Start'
When that happens I'm getting this in the console:
```
12:58:55 GMT+0000 (GMT) ERROR errorHandler: ReferenceError: domain is not defined
	 startActiveScan    https://zap//zapCallBackUrl/400785326767206905?name=tools/activeScan.js 115:2: {} 400785326767206905:682:35

PR #351, opened 2019-01-11
Refactor/Fix Startup

### LINKED ISSUE #326: Tutorial 'Dont show this screen again' option broken

### LINKED ISSUE #342: HUD unresponsive if you wait on the welcome screen
Launch a browser with the welcome screen enabled, and leave it for a couple of mins.
Dismiss the screen and you'll find that the HUD doesnt communicate with ZAP - no alerts etc.
It looks like the hearbeats are failing because navigator.serviceWorker.controller is null.
@dscrobonia is this because the serviceworker isnt really active until we refresh?
If so we might need to change how the Welcome screen works, unless your startup revamp will address this?

dscrobonia:
The iframes are now individually refreshed, instead of the entire target page, allowing the service worker to take control of them without a painful startup experience.

The startup still feels a little abrupt to me, and I'd like to figure out how to fade the panels in, but gave up after a few minutes trying on that. 

This should help with a decent amount of startup issues.

Give a try and let me know how it works.

Fixes #326, fixes #342 

- [ ] Add documentation about how startup works


dscrobonia:
Just fixed some of the startup bugs and added a timer to print out how long the delay is between when the management script gets loaded until the UI is shown. I ran it roughly a dozen times and my startup times on FF were roughly 2s - 3.5 seconds on average. that's not too bad, but still seems like a longtime to be waiting for some indication that something is happening.

I think during that period of time would be a great use of adding a "loading screen" - something like what I tried before here: 
image

This loading screen would get shown while the HUD loaded, and then the "welcome screen" would get displayed after that.

Let me know what y'all think!

psiinon:
About 1/2 of the time the alert tools are not getting updated with the counts for me until you go to another page - I can see the growl events but the counts stay at zero.
Almost certainly a timing issue, but one we'll need to sort out, first impressions and all that..

psiinon:
@dscrobonia I've just pushed 2 commits which:

1. Merge the latest changes
1. Update the tests so they are hopefully passing now

psiinon:
Removed [WIP] from the title based on gchat with @dscrobonia

psiinon:
Just fixed the conflict and re-enabled the 'dont show this again' option.
So far looking good :)

psiinon:
Looks like this is close but its still failing the tests too often. I think I'm getting close to the problem, and its looking like a timing issue in the HUD. Will be a good one to fix...

psiinon:
Fixed what I'm hoping is the problem, thanks to help from @thc202 , fingers crossed for the tests...

psiinon:
Updated first comment as this fixes #326 and #342

dscrobonia:
Everything seems fine except that after click the "dont show this again" check box and then closing the browser and relaunching it I still see the popup. Am i the only one expeirence that/ was this covered somewhere else?

thc202:
You need a new version of WebSockets add-on that has zaproxy/zap-extensions#1927 (it was not yet included in the weekly).

src/test/java/org/zaproxy/zap/extension/hud/ui/uimap/HUD.java
```diff
@@ -126,10 +126,11 @@ public WebElement waitForDisplayPanel() {
         return null;
     }
 
-    public List<WebElement> waitForHudButtons(int expected) {
+    public List<WebElement> waitForHudButtons(By byPanel, /*WebElement panel, */ int expected) {
```
thc202: Was the `panel` added on purpose?

src/test/java/org/zaproxy/zap/extension/hud/ui/uimap/HUD.java
```diff
@@ -157,13 +158,6 @@ public WebElement waitForDisplayPanel() {
 
     public void openUrlWaitForHud(String url) {
         this.webdriver.get(url);
-
-        if (hudNotInstalled) {
```
thc202: The `hudNotInstalled` should be removed (no longer used).

psiinon:
Fixed conflicts and comments

thc202:
Thanks!

PR #352, opened 2019-01-11
Active scan specific request

### LINKED ISSUE #299: Scan specific url
Hi all -

I started using HUD with ZAP and it has been a great experience. Very efficient way to do the manual testing. My request is, it would be great to give a button that can trigger an active scan on the intercepted request. 

Thanks!
Nagpradis

psiinon:
Fixes #299

Be good to get some feedback on this, in particular the UI.
This adds an 'Active Scan' button to the standard HTTP Request/Resp dialog.
The button is only enabled if the URL is in scope and theres no other active scan going on, but theres no 'Are you sure?' dialog.
It works most of the time but sometimes the scan progress (in the standard active scan control) stays at zero :(
Not sure why, will need to investigate.


psiinon:
This pull request **introduces 3 alerts** when merging bf69758f96402c941520597ab4943b71a64d5b31 into 90986ba51db2bbe580f6c24cc6ec51efd76b9e7b - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-hud/rev/pr-efa4d9d049f331a4d51ac9e0cc9ab9243ec0336e)

**new alerts:**

* 3 for Unused variable, import, function or class

---

dscrobonia:
This worked and looked great to me! :smile: :+1: Love the UI! Great use of disabling it when not in scope. In the future if we continue to build on this pattern and add tools to the http message panel (which I think we should, its awesome) then we can decide then if we want switch from text + icon on the button to just the icon. Overall great add though!

I didn't notice any issues with the progress. lmk if you can get repro steps, would love to debug it.

I have some questions about the code, but it may just be dev left overs that need cleaning up.

src/main/zapHomeFiles/hud/tools/activeScan.js
```diff
@@ -87,12 +87,43 @@ var ActiveScan = (function() {
...
 			.then(upgradedDomain => {
-				return utils.zapApiCall("/ascan/action/scan/?url=" + upgradedDomain)
-			}).
-			then(response => {
+				startActiveScan(tabId, upgradedDomain, "true", "GET", "");
+			})
+			.catch(utils.errorHandler);
+	}
+	
+	function showReqDialog(tabId, uri, method, body) {
```
dscrobonia: what is the purpose of this function? I noticed it wasn't called anywhere.

src/main/zapHomeFiles/hud/tools/activeScan.js
```diff
@@ -87,12 +87,43 @@ var ActiveScan = (function() {
...
+				return config;
+			})
+			.then(config => {
+				return utils.messageFrame(tabId, "display", {action:"showDialog", config:config})
+			})
+			.catch(utils.errorHandler);
+	}
+
+
+	function startActiveScan(tabId, uri, recurse, method, body) {
```
dscrobonia: I'd prefer to keep this method as simple as needed. For example, do we have a need for changing the method? or adding a body? I think we do need to differentiate between scanning a url vs scanning a domain, but that could either be handled by two functions `startActiveScanDomain` vs `startActiveScanUri` and those should both only expose the parameters they need for now. Realistically though I think we could get away with a single function `startActiveScan(tabId, uri, recurse)` which would allow us to support both use cases in a single function with just the parameters we need.

src/main/zapHomeFiles/hud/tools/activeScan.js
psiinon: Ah yeah, I did expetiment with adding a dialog to confirm that an ascan should be started, but in the end didnt use it. Will delete.

src/main/zapHomeFiles/hud/tools/activeScan.js
psiinon: Yes we do need these - they are needed (and specified) when scanning a POST request. If they're not provided then we'll end up scanning a GET request instead, which isnt what we want to do.

psiinon:
Screenshot:
ascan-url

psiinon:
Interestingly I cant reproduce the issue on my linux box. Will try again with the windows one..

psiinon:
This pull request **introduces 1 alert** when merging a8dae543a82d9a29eed8fbbff437a3a5a84062ca into 8cae68d0ddb27a0f8d3099fb2691afe04462e923 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-hud/rev/pr-daf5cc109598f971f33ec3727e8fa43babf96c5b)

**new alerts:**

* 1 for Syntax error

---

psiinon:
@dscrobonia ok to merge this now?

src/main/zapHomeFiles/hud/tools/activeScan.js
dscrobonia: ah gotcha. I see it now. thanks!

src/main/zapHomeFiles/hud/tools/activeScan.js
```diff
@@ -40,7 +40,7 @@ var ActiveScan = (function() {
 		registerForZapEvents(ACTIVE_SCAN_EVENT);
 	}
 
-	function showDialog(tabId, domain) {
+	function showStdDialog(tabId, domain) {
```
dscrobonia: Sorry to leave once last nit pick. Was there a reasn to change this function name? Across almost all of the other tools we use just `showDialog` would be nice to leave it the same if possibe.

psiinon: Yeah, its because I introduced a new dialog for scanning just this request but then later removed it.
Will change back to the old one :)

psiinon: Done

PR #356, opened 2019-01-16
[WIP] Use specific TLS version in tests

thc202:
To not break the tests in Java 11 while the problem is being
investigated.


thc202:
For the record, this is "caused" by zaproxy/zaproxy#5186, ZAP is no longer hardcoding the enabled SSL/TLS versions so version 1.3 is also enabled by default which is causing some HUD requests to fail.

psiinon:
Be good to get another review of this so we can merge it.
It fixes the issue that causes the Java 11 tests to fail, without it we wont be able to merge anything else :/

PR #371, opened 2019-01-28
Update vuejs-i18n and package lock file

psiinon:
This is based on #362 (because I like the updates to package-lock.js but
also actually updates the file in the right directory.
Basic testing indicates to problems.
Oh, and the vuejs-i18n version is wrong as per https://github.com/kazupon/vue-i18n/issues/488#issuecomment-458110274

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -97,7 +97,7 @@
         },
         "minimist": {
           "version": "1.2.0",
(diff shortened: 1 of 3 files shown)


psiinon:
Closing in favour of #362 which I've updated (good suggestion @thc202 ;)

Issue #344, opened 2019-01-07
Make functional tests more reliable


thc202:
Some of the functional tests (tutorial) fail sometimes, e.g.:
```
org.zaproxy.zap.extension.hud.ui.firefox.badsite.BadSiteUnitTest > cannotAccessApiKeyWhenLoadingUtilsAsScript(FirefoxDriver) FAILED
    org.openqa.selenium.JavascriptException: ReferenceError: LOG_ERROR is not defined
    Build info: version: '3.13.0', revision: '2f0d292', time: '2018-06-25T15:24:21.231Z'
    System info: host: 'travis-job-acbae740-7b82-4d89-9d65-e682edd3a1f2', ip: '127.0.1.1', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-101-generic', java.version: '1.8.0_151'
    Driver info: org.openqa.selenium.firefox.FirefoxDriver
    Capabilities {acceptInsecureCerts: true, browserName: firefox, browserVersion: 64.0, javascriptEnabled: true, moz:accessibilityChecks: false, moz:geckodriverVersion: 0.23.0, moz:headless: true, moz:processID: 5738, moz:profile: /tmp/rust_mozprofile.RP2aKE..., moz:shutdownTimeout: 60000, moz:useNonSpecCompliantPointerOrigin: false, moz:webdriverClick: true, pageLoadStrategy: normal, platform: LINUX, platformName: LINUX, platformVersion: 4.4.0-101-generic, rotatable: false, setWindowRect: true, timeouts: {implicit: 0, pageLoad: 300000, script: 30000}, unhandledPromptBehavior: dismiss and notify}
    Session ID: ad53072a-f6f6-4e5e-a008-bf11c07ff06c
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at org.openqa.selenium.remote.http.W3CHttpResponseCodec.createException(W3CHttpResponseCodec.java:187)
        at org.openqa.selenium.remote.http.W3CHttpResponseCodec.decode(W3CHttpResponseCodec.java:122)
[...]

PR #376, opened 2019-01-29
Load latest upgraded domains

### LINKED ISSUE #162: https upgrade bug on startup
Everytime i restart the HUD the https upgrade has a bug on the first load. I have to refresh the tab to have it work. The `sharedData.upgradedDomains` doesn't include the target page. This means that scope, alerts, history, and any other tool taht depends on the correct protocol for the api won't work

I believe this is because the domain is only added when ZAP sends the event via websockets. But on the first load the event is sent before the HUD can receive it (?). Probably should wrap this into a more robust startup, which I should focus on in #160 . This may require a more thorough description of how startup works.

@psiinon

thc202:
Change the HUD (`serviceworker.js`) to load the upgraded domains from
ZAP after establishing the WebSocket connection to ensure the HUD knows
about them from the beginning, if not the HUD might miss the event that
notifies of the upgrade.
Remove (empty) initialisation of `upgradedDomains` in `utils.js` now
being done in `serviceworker.js`.
Extract a method in `ExtensionHUD` that normalises the upgraded domain.
Add a view to `HudAPI` that exposes the upgraded domains.
Update changelog.

Fix #162 - https upgrade bug on startup
Related to #344 - Make functional tests more reliable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Add tasks for Enable, Show and Break tutorial pages
  - Add plain text/regex filtering capability to History section [#286](https://github.com/zaproxy/zap-hud/issues/233)
 
+### Fixed
+ - Correct handling of upgraded domains on startup. [#162](https://github.com/zaproxy/zap-hud/issues/162)
+
 ## [0.2.0] - 2018-12-31
 
 ### Added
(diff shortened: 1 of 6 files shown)


psiinon:
I'm still getting some failure locally, but if it passes consistently on travis then I'll approve it as its clearly better than what we have now. It passed the first travis run and have restarted it to double check.

psiinon:
My local tests are passing consistently now - just needed to update to the latest weekly version (thanks @thc202 ;)

PR #377, opened 2019-01-29
Stop the tutorial server on unload

thc202:
Change ExtensionHUD to stop the server when unloaded, otherwise the
add-on is not properly uninstalled.
Update the changelog.


psiinon:
Now conflicting

thc202:
Addressed.

PR #379, opened 2019-01-30
Add show comments tool

### LINKED ISSUE #378: Show comments tool
Add a tool to show a count of the number of comments and to allow them to be displayed, similar to the showHide tool.
There are various options for showing the comments:

1. Changing them into text nodes
1. Inserting an image that shows the comment on hover over or on click
1. Showing the comments as a list in the bottom tab

psiinon:
Fixes #378

When turned on this injects icons where the comments occur with the
title set to the comment test, so that you can see it when you hover
over the icon.
This works well for simple pages but doesnt work so well for complex
sites like amazon.com or bbc.co.uk
Thoughts?


psiinon:
Fixed the problem with comment icons not appearing on some sites.
However if you load bbc.co.uk with comments on then it hangs the browser :( Works fine if you turn the comments tool on after the page has loaded. That will need to be fixed.
I think the UI is ok, but be good to get some feedback on that.
One nice enhancement I've thought of is a (user up-datable) list of key words wich if present cause a red icon to be shown (eg 'password', 'admin' ;)

dscrobonia:
Didn't have any luck using this. :/ The comment counter didn't seem to work, and when i used the tool the comment icons would appear on the target page, but there was no way to view the comments.

psiinon:
Hover your mouse over the image and the comment should appear.
Which sites was the counter not working with?

dscrobonia:
Would love to get this one through soon! A few notes:
* the tool doesn't update the count when you first turn it on, it only update the count when loading a new page
* we should not count comments that are only whitespace (weird, but they do exist)
* we need to figure out error handling for pages that break.

psiinon:
Updated to latest code, plan to address the issues soon...

dscrobonia:
i actually had somebody reach out to me asking where that awesome comment feature they saw me demo was. ;) I'd be in favor of getting out a less than perfect version of this tool and then iterate from there.

psiinon:
OK, I think this is ready for review.
I've changed it so that it ignores comments just containing whitespace and added a tutorial page with associated task.
It seems to be working properly now on pages that it was having problems with, like amazon.com and bbc.co.uk but I need to try more sites.
No unit tests yet - I'm aiming on adding those, and would still like to flag suspicious comments with a red flag, but thats a nice to have.

psiinon:
Just changed the tool locally to flag comments with suspicious content, based on https://github.com/zaproxy/zap-extensions/blob/master/addOns/pscanrulesBeta/src/main/zapHomeFiles/xml/suspicious-comments.txt
2 issues - I'm not sure the exclamation on the balloon is that obvious:
sus icious-comments
should the exclamation be in red, or the whole comment icon?
Also what should the tool status show?
My current thought is that it should just change to the new icon (once decided) to indicate at least one comment has suspicious content but to keep showing the total number of comments.
Alternatively it could change to just showing the number of comments with suspicious content, or both, eg 1 / 3 which would indicate there are 3 comments, one of which has suspicious content. 
Thoughts?

psiinon:
Is this icon better?
balloon--exclamation2

kingthorin:
The yellow on yellow doesn’t work well. Could the caution sign part be made more orange’ish? 

I like the idea of keeping a cumulative count but changing the icon for suspicious comments. The count could become unruly on some pages.... think of having two three digit numbers or something like: 123/456.

Are you able to leverage the pscan file or did the content have to be duplicated? (Just wondering from a maintenance perspective.)

psiinon:
Another alternative :)
balloon-red

Re maintenance, its not an exact copy as I'm just checking for 'contains' rather than word matches, and the plan is for this to be in the HUD configs so that users can change the strings to be any that they like, via a right-click option in the HUD.

kingthorin:
Oh I like that. Yellow when there are comments, red/pink when there are suspicious strings.

Re maint, sounds good to me.

kingthorin:
Can we give users a way to mark something as “Not suspicious”? (Maybe I’m thinking a step too far for the initial impl.)

psiinon:
For that they can just remove the keyword and refresh the page :) not sure what else could be usefully done...

psiinon:
Ugh, realised we might actually 4 different icons now for the tool state:

- Tool off, no suspicious comments (white balloon)
- Tool off, suspicious comments
- Tool on, no suspicious comments (yellow balloon)
- Tool on, suspicious comments

Maybe the red exclamation icon as per https://github.com/zaproxy/zap-hud/pull/379#issuecomment-521200919 on both the white and yellow balloons would be better?

psiinon:
I've just realised theres a problem with this PR - the tool status is shared across tabs - it should be tab specific.
You probably wont notice but it becomes much more apparent when suspicious comments are flagged ;)
Will fix with the change to support (hardcoded) suspicious comments...

psiinon:
Updated the PR to flag (hardcoded) suspicious strings and to fix the problems with multitab.
However its not ready to merge yet.
The tool needs to have per tab state, as different tabs can have different numbers of comments. To make that work I included the tabid in the localforage config.
Unfortunately tabs get a new id everytime a page is loaded, this means the tool is always turned off, and we'll also leak storage.
@dscrobonia (or anyone else) any thoughts on how to fix that?
The tool status also doesnt get updated the first time the tool is added - will need to address that too.
Other than that I think its looking good - feedback appreciated.
The plan is still to allow the user to change the string used to flag suspicious comments.

kingthorin:
Store a hash of the comments and update the tabID if the hashes match? 
Yes you could have multiple tabs of the same page but if the comments match and you display the same info who cares?

psiinon:
I was wondering about using the url as we do for issue flags, so thats similar. But the tool state also stores whether the tool is on or off. Wouldnt it be weird to go to a new page and for the tool to turn on (or off) automatically just because its open in another tab?

kingthorin:
I dunno I guess there’s an edge case where you have the same resource displaying in two tabs and want comments tool enabled in one but disabled in the other, but I do think it’s an edge case.

psiinon:
I'm not so sure - I think it could be one of those 'annoyances' that could be irritating.
 I can also see us having more tools in the future that are tab specific, so I'd really like to have a standard solution for this if we can...

psiinon:
A brief chat with @dscrobonia this morning helped me realise that the comments tool doesnt actually need to persist its state - I've stopped it doing that and it now seems to work as intended!
I still plan to add tests and to make the suspicious strings users configurable, but I think its fine to review and hopefully merge this PR before those changes.

thc202:
(From talks in IRC.)

Improvement suggestions:
 - Allow to easily copy the contents of the comment, e.g. by clicking the icon;
 - Indicate (e.g. highlight) which word in the comment is suspicious (sometimes the comment is big and it's not immediately apparent which word it is).

When adding the tool the count/icon does not reflect the actual state.

psiinon:
The show/enable tool also isnt updated when added to a page, so I've raised an issue for that https://github.com/zaproxy/zap-hud/issues/590
I've updated the tool and help so that clicking on the icon copies the text to the clipboard, but I cant see how to highlight the suspicious term - thoughts? I could use underscores around the term but that then changes it, and markup doesnt seem to get applied.

thc202:
No suggestions on how to do that but for me it can be added in a follow up PR.

dscrobonia:
re: Per Tab Storage I had an idea as I'm not sure storing them in memory as variables will work as desired (see: https://github.com/zaproxy/zap-hud/issues/590#issuecomment-526654177)

## Background
The way we track tabs currently is by extracting the **tabId** off of the URL param of a HUD frame.
i.e.
https://zap//zapCallBackUrl/8539492085811379914/file/management.html?frameId=management& **tabId=5549153-51**

The URL of each running tab is accessed using the [Clients API](https://developer.mozilla.org/en-US/docs/Web/API/Clients). An example of this in use can be [seen here](https://github.com/zaproxy/zap-hud/blob/develop/src/main/zapHomeFiles/hud/utils.js#L435):
```
function messageFrame(tabId, frameId, message) {
		return clients.matchAll({includeUncontrolled: true})
			.then(clients => {
				for (let i = 0; i < clients.length; i++) {
					let client = clients[i];
					let params = new URL(client.url).searchParams;
	
					let tid = params.get('tabId');
					let fid = params.get('frameId');
	
					if (tid == tabId && fid == frameId) {
						return client;
					}
				};
	
				throw new NoClientIdError('Could not find a ClientId for tabId: ' + tabId + ', frameId: ' + frameId);
			})
```

The beauty of using the Clients API is that we don't need to manage anything to track active tabs. The downside is we don't have a storage abstraction to build upon.

## Suggestion
To store per tab data for a tool we should store the data indexed by the tabId, and then occasionally "clean up" the per tab store by checking which tabId's being stored are still active in the clients API. What would make this cleaner is abstracting it out to something we can re-use for all tools.

So we would have a `TabStore` that would just wrap around indexeddb. The `TabStore` would be index objects by toolId, and then within those objects by tabId. The `TabStore` interface would be a simple read/write like `localforage` provides for indexeddb. 

But the catch is that every time a new tab is opened `TabStore` will automatically compare what tabId are being stored vs what is considered an active tab in the Clients API and then **delete any stored data for expired tabs**. This will help us prevent memory leaking, while also allowing per tab data.

psiinon:
@dscrobonia I cant reproduce the problems you've seen on chrome - it looks like its working for me with multiple tabs on both Firefox and Chrome. Can you give a reproducible failing example?
Merged latest changes so no conflicts right now.

psiinon:
Ugh, so pushing from a windows box wasnt such a good idea :/ will fix the spotlessApply when I'm back...

thc202:
It's actually failing JS linting (of `showComments.js`).

psiinon:
Ah ha! I just saw so loads of errors and jumped to conclusions. OK, I'll work on those...

psiinon:
@thc202 is there an easy way to run xo --fix via gradle? I've tried adding 'fix', '--fix', 'lint-fix' to build.gradle.kts but no joy. Still on the train so limited connectivity :/

thc202:
The task `npmLintStagedHud` runs `xo --fix` (it also runs `ava` and `git add` though).

Issue #345, opened 2019-01-07
Run functional tests with Chrome


thc202:
To ensure that the HUD also works correctly in Chrome (it should). This depends on a new release of [selenium-jupiter](https://github.com/bonigarcia/selenium-jupiter) to easily test both of them (Firefox and Chrome) using the same code.

thc202:
The newer version (3.0.0) is already available in Maven Central.

PR #383, opened 2019-01-31
Change tests to run with multiple browsers

thc202:
Change `BrowsersTest` to include Chrome (commented until the tests are
more reliable #344), also, use `proxy` instead of manually setting the
proxy preferences to Firefox.
Update all browser tests to run the tests with multiple browsers (by
using `TestTemplate`).
Change `testTaskAndNextButton` in `AlertNotificationsPageUnitTest` and
`PageAlertsPageUnitTest` to reset the tasks' state to have the task in
the expected state for each browser.
Tweak exception message asserted in `BadSiteUnitTest` to work with
multiple browsers (the message prefix is different).
Update Selenium-Jupiter, required for testing with multiple browsers.
Remove test home always, to run the tests with fresh ZAP.

Part of #345 - Run functional tests with Chrome

---
Old description:

Initial changes for #345.

The (tutorial) test classes are still in the `firefox` package and the base class is still named after Firefox. (That will be changed once we decide how we want the tests to be executed by the browsers.)
Currently each test method is being executed sequentially by the browsers (Firefox then Chrome) which causes some tests to fail as the preconditions are no longer met (i.e. tutorial task is already completed). Should all tests be executed by one browser, ZAP reset and then the other browser? Or, change the test methods to ensure the preconditions are the expected for each browser?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -131,7 +131,7 @@ dependencies {
     testImplementation("org.junit.jupiter:junit-jupiter-params:$jupiterVersion")
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:$jupiterVersion")
 
-    testImplementation("io.github.bonigarcia:selenium-jupiter:2.2.0")
+    testImplementation("io.github.bonigarcia:selenium-jupiter:3.3.0")
     testImplementation("org.hamcrest:hamcrest-all:1.3")
     testImplementation("org.mockito:mockito-all:1.10.8")
     testImplementation(files(fileTree("lib").files))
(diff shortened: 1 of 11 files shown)


psiinon:
Good work!
We already have a resetTutorialTasks API action - could we use that?
If not are there other API calls that would help?
I like the idea of resetting anything that needs resetting as part of the tests so that we dont need to keep restarting ZAP.

thc202:
Tests being relocated in #597.

> We already have a resetTutorialTasks API action - could we use that?

Yes, that should work.

thc202:
No longer WIP.

thc202:
WIP again, with Chrome #344 is more noticeable. I'll take another look at that.

thc202:
Disabled Chrome for now (per #344), should be ready to merge.

PR #396, opened 2019-02-06
Link to ZAP HUD group

psiinon:
From tutorial and a new desktop Online menu item


psiinon:
We already refer to the std ZAP User Group as such, so I was concerned there would be confusion there. I was also thinking this would be both for HUD users and developers, as I'm hoping we can make it sufficiently easy to tweak the HUD for there to be more overlap between the 2 than we see with ZAP...

psiinon:
Just added an update to the changelog.

PR #398, opened 2019-02-06
Use org.zaproxy.add-on plugin

thc202:
Add the ZAP plugin to build add-ons and update the build accordingly.
The following tasks were renamed:
 - `assembleZapAddOn` to `jarZapAddOn`;
 - `copyAddOn` to `copyZapAddOn`;
 - `deploy` to `deployZapAddOn`;
 - `installAddOn` to `installZapAddOn`;
 - `uninstallAddOn` to `uninstallZapAddOn`.

The add-on will be built into the directory `build/zapAddOn/bin/`
instead of `build/zap/`.
Update Gradle to 5.2.1, plugin requirement.
Remove local plugin, extensions, and tasks which are now provided by the
added plugin.
Remove `ZapAddOn.xml`, it will be generated by the plugin.


thc202:
The build will fail, Gradle plugin is not yet published.

thc202:
Addressed the conflicts. The build is now passing, pushed a (temporary) commit to use the Gradle plugin directly from the repo, to reduce the work needed to try/review the changes.

Issue #199, opened 2018-10-01
Multiple Serviceworkers in Isolated Tabs


dscrobonia:
Testing with Firefox 62 it appears that instead of a single service worker process shared among all tabs (per domain) there is now a service worker proccess **per tab** (per domain). Loading new pages within a single tab still keep the same service worker, but openning a new tab spawns a new service worker.

I believe this change still complies with the service worker spec, which defines service workers as stateless, but has broken our model of how we use service workers because of a few hacks we have made to keep the serviceworker performant. This will require some changes to how I had originally mapped out multi tab support, but should still be doable.

I do still need to test on chrome. If the firefox and chrome implementations of service workers has diverged (which I would guess they have) we'll have to consider that as well.

psiinon:
So https://bugzilla.mozilla.org/show_bug.cgi?id=1438945 says its fixed in Firefox 65 - we should double check to see if this is all ok now...

dscrobonia:
How exciting! I'll need to investigate...

PR #401, opened 2019-02-07
Persist the UI configs to ZAP

### LINKED ISSUE #321: Persist UI options in the ZAP configs
This is so that UI options are not lost when browser launch is used.
Depends on #319

psiinon:
Fixes #321 

The tools are now read from the ZAP configs via the API - if they are not set then they are defaulted to the values now held in serviceworker.js
I suspect this is slowing down start up a little bit but I dont have any measurements to prove it.
Injecting the default values into serviceworker.js would probably involve fewer changes and be quicker, but I thought I'd try the 'right' way first :)
I've seen a couple of cases where the tools are duplicated - this looks like its a client side issue as the values stored in the ZAP config file have been correct. I'm guessing its a timing issue when loading the tools but will need to check that.
General feedback appreciated - I'll look into the above bug asap.
It might also be worth waiting until #351 is merged before merging this one.

### DIFF
--- a/src/main/zapHomeFiles/hud/serviceworker.js
+++ b/src/main/zapHomeFiles/hud/serviceworker.js
@@ -3,6 +3,9 @@ var ZAP_HUD_FILES = '<<ZAP_HUD_FILES>>';
 var toolScripts = [
 	'<<ZAP_HUD_TOOLS>>'
 ];
+// The configured tools
+var CONFIG_TOOLS_LEFT = '<<ZAP_HUD_CONFIG_TOOLS_LEFT>>';
+var CONFIG_TOOLS_RIGHT = '<<ZAP_HUD_CONFIG_TOOLS_RIGHT>>';
 
 importScripts(ZAP_HUD_FILES + "?name=libraries/localforage.min.js"); 
 importScripts(ZAP_HUD_FILES + "?name=libraries/vue.js"); 
@@ -56,9 +59,61 @@ localforage.setItem("tools", [])
 	})
 	.catch(utils.errorHandler);
 
-const onInstall = event => {
-	utils.log(LOG_INFO, 'serviceworker.install', 'Installing...');
+initWebSockets();
 
+function initWebSockets() {
+	/* Set up WebSockets */
+	let ZAP_HUD_WS = '<<ZAP_HUD_WS>>';
+	webSocket = new WebSocket(ZAP_HUD_WS);
+
+	webSocket.onopen = function (event) {
+		// Wont be able to log until ther websocket is set up
+		utils.log(LOG_INFO, 'serviceworker.initWebSockets', 'Initialized');
+		// Basic test
+		webSocket.send('{ "component" : "core", "type" : "view", "name" : "version" }'); 
+
+		apiCallWithResponse("hud", "view", "upgradedDomains")
+		.then(response => {
+			let upgradedDomains = {};
+
+			for (const domain of response.upgradedDomains) {
+				upgradedDomains[domain] = true;
+			}
+			return localforage.setItem('upgradedDomains', upgradedDomains);
+		})
+		.catch(utils.errorHandler);
+	};
+
+	webSocket.onmessage = function (event) {
+		// Rebroadcast for the tools to pick up
+		let jevent = JSON.parse(event.data);
+
+		if ('event.publisher' in jevent) {
+			utils.log(LOG_DEBUG, 'serviceworker.webSocket.onmessage', jevent['event.publisher']);
+			var ev = new CustomEvent(jevent['event.publisher'], {detail: jevent});
+			self.dispatchEvent(ev);
+		} else if ('id' in jevent && 'response' in jevent) {
+			let pFunctions = webSocketCallbacks[jevent['id']];
+			let response = jevent['response'];
+			if ('code' in response && 'message' in response) {
+				// These always indicate a failure
+				let error = new Error(I18n.t("error_with_message", [response['message']]));
+				error.response = response;
+
+				pFunctions.reject(error);
+			} else {
+				pFunctions.resolve(response);
+			}
+			delete webSocketCallbacks[jevent['id']];
+		}
+	};
+
+	webSocket.onerror = function (event) {
+		utils.log(LOG_ERROR, 'websocket', '', event);
+	};
+};
+
+const onInstall = event => {
 	// Cache Files
 	// not sure caching in service worker provides advantage over browser - may be able to remove
 	event.waitUntil(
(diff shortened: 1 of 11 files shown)


psiinon:
OK, this is now working nearly all of the time.
About 1 in 6 times (for me) the left frame fails to load.
If you put debug logging on utils.loadFrame and utils.saveFrame you'll see its because on startup the left panel load frame (called from panel.js via utils.loadPanelTools) happens before the save frame (called from utils.initializeHUD)
@dscrobonia (or anyone else) any suggestions?
I guess this could have happened without my changes, but the delay introduced by calling ZAP to read the configs just makes it a little bit more likely to happen.

psiinon:
The easiest way to reproduce the problem is to just keep launching Firefox from ZAP until the left frame fails to load.

psiinon:
After some playing around I've changed it so that the default tools are injected.
This seems to have fixed the problem and should speed up the initial load ever so slightly :)
I think its now ready :)

psiinon:
As @thc202 has pointed out, this is actually broken.
Dont bother reviewing it until I've fixed it :/

psiinon:
Have fixed most of the problems with this, but just spotted another one - once the Reset option is used all of the tools appear in the 'add tool' dialog - will look into that now...

psiinon:
Arg, not quite there yet - resetToDefaults is only working in the first tab :/

psiinon:
The problem is that with Firefox the webSocket is null for all except the first tab, as per #199 - it works fine in chrome.
I'll see if I can work out a solution...

psiinon:
I'm struggling with this one :/
@dscrobonia can you think of a way to communicate to the first service worker or will we just need to say that on Firefox you can only change the configs in the first tab until that Fx bug is fixed?

psiinon:
Just pushed a fix which prevents the tools re-registering for ZAP events, which could have significantly increased the number of messages received. Good spot @thc202 :)

psiinon:
Think I've just fixed the problem on Firefox :)
The issue was that I'd moved the websocket initialization into the onInstall handler to reduce problems with making the upgradedDomains API call too early.
The problem is that onInstall appears to only be called for the first tab.
Moving the key websocket initialization code back inline has fixed this.
Hoping this is good to go now!

dscrobonia:
Just tested it out and no bugs! Woohoo! I'm running low on sleeps tho so I'll do the code review tomorrow morn. :p Excited for this one!

src/main/zapHomeFiles/hud/serviceworker.js
```diff
@@ -56,9 +59,9 @@ localforage.setItem("tools", [])
 	})
 	.catch(utils.errorHandler);
 
-const onInstall = event => {
-	utils.log(LOG_INFO, 'serviceworker.install', 'Installing...');
+initWebSockets();
```
dscrobonia: does the placement of this call above `onInstall` with the function defintion below it matter? From my understanding it shouldn't, but thought I'd check. It would be nice if `onInstall` and `onActivate` were next to each other because they're related functions.

src/main/zapHomeFiles/hud/utils.js
```diff
@@ -246,7 +243,20 @@ var utils = (function() {
 
 		return Promise.all(promises)
 	}
-	
+
+	/*
+	 * Backs up the frames tools to ZAP
```
dscrobonia: could we add a little more context in this comment about what this means? Reading "back up" wasn't intuitive to me, so it may be helpful to describe that were persisting configs through multiple loads of the HUD

src/main/zapHomeFiles/hud/utils.js
```diff
@@ -246,7 +243,20 @@ var utils = (function() {
...
+	 */
+	function backupFrame(frame) {
```
dscrobonia: Likewise hear I think it would be more clear if the function was named `backupFrameTools` seeing as thats what were saving.

src/main/zapHomeFiles/hud/serviceworker.js
psiinon: No, but can easily move it.

src/main/zapHomeFiles/hud/serviceworker.js
```diff
@@ -155,72 +210,31 @@ const logHandler = event => {
...
-};
+self.addEventListener('hud.backup', backupHandler);
 
 function registerForZapEvents(publisher) {
 	apiCall("event", "register", publisher);
 };
 
 function apiCall(component, type, name, params) {
+	if (!webSocket) {
+		console.log("serviceworker.apiCall no WebSocket " + component + " " + type + " " + name);
```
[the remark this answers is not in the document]
psiinon: No, that would then call apiCall and we'd get in a loop ;) Can add a comment to that effect.

src/main/zapHomeFiles/hud/serviceworker.js
```diff
@@ -288,17 +306,28 @@ function showHudSettings(tabId) {
...
-		.then(utils.loadAllTools)
+	utils.loadAllTools()
 		.then(tools => {
 			var promises = [];
+			// Clearing these values means they will go back to the default when next read
+			promises.push(backup('leftPanel', ''));
+			promises.push(backup('rightPanel', ''));
 
 			for (var tool in tools) {
-				promises.push(self.tools[tools[tool].name].initialize());
```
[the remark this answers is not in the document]
psiinon: Yes, because the initialize() functions dont return a promise, and should therefore not be treated as such.

src/main/zapHomeFiles/hud/utils.js
psiinon: I'd rather not change this, as we might well back up other frame related data such as the frame position in the future.

psiinon:
@dscrobonia think I've addressed all of those comments with code comments or replies :)

src/main/zapHomeFiles/hud/serviceworker.js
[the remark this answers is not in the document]
dscrobonia: haha. got it!

src/main/zapHomeFiles/hud/serviceworker.js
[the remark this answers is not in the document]
dscrobonia: ooh good catch! :D 

we could change the initialize() functions to return utils.writeTool so that they return a promise, allowing us to ensure that these steps get completed, but thats not necessary if there's nothing that follows thats dependent on the tools being initialized. 

I'm a little concerned that there might be something that is dependent on them so by letting them run async we might create a hard to find race condition. 

Thoughts?

src/main/zapHomeFiles/hud/utils.js
dscrobonia: That make sense. Then maybe just in the comment above describing what "backup" means would be good.

dscrobonia: ah you did that already!

psiinon:
Yes we should, but I think we should do that in another PR. I think there are various other places where we should be returning promises but are not :)

psiinon:
Raised #438 for that :)

PR #404, opened 2019-02-08
Added -hud ZAP command line option

psiinon:


src/main/java/org/zaproxy/zap/extension/hud/launch/ExtensionHUDlaunch.java
```diff
@@ -0,0 +1,179 @@
...
+            argClasses[i] = arg.getClass();
+            i++;
+        }
+        Method method = obj.getClass().getMethod(methodName, argClasses);
+        return method.invoke(obj, args);
+    }
+
+    private void launchBrowser() {
+        new Thread(
+                        new Runnable() {
```
thc202: Could use a lambda.

src/main/java/org/zaproxy/zap/extension/hud/launch/ExtensionHUDlaunch.java
```diff
@@ -0,0 +1,179 @@
...
+                                        getExtHUD().setHudCmdlineOptionUsed(true);
+                                        getExtHUD().getHudParam().getConfig().save();
+                                    }
+                                    // Object wd =
+                                    callMethodByReflection(
+                                            getExtSelenium(), "getProxiedBrowserByName", "Firefox");
+
+                                    // callMethodByReflection(wd, "get", "https://zaproxy.org");
+                                } catch (Exception e1) {
+                                    if (!View.isInitialised()) {
```
thc202: Log always? Should ZAP exit if not able to start the browser when in daemon?

src/main/java/org/zaproxy/zap/extension/hud/launch/ExtensionHUDlaunch.java
```diff
@@ -0,0 +1,179 @@
...
+                        },
+                        "ZAP-BrowserLauncher")
+                .start();
+    }
+
+    @Override
+    public void execute(CommandLineArgument[] args) {
+        if (arguments[ARG_BROWSER_IDX].isEnabled()) {
+            this.launchBrowser();
+        } else {
```
thc202: The else can be removed.

src/main/java/org/zaproxy/zap/extension/hud/launch/ExtensionHUDlaunch.java
psiinon: I think exiting is a bit too harsh in this case

psiinon:
Think that should address all of the comments

thc202:
Thanks!

PR #414, opened 2019-02-17
add transitions to buttons

dscrobonia:
This adds transitions to the buttons when you hover over them to make them slide open. 

The panels will wait until all buttons have closed before shrinking the iframe back down. (add a border to the iframe if you're curious about this)

slidey-buttons


thc202:
Worth updating the changelog.

PR #418, opened 2019-02-18
Wait for frames and its state in frames test

thc202:
Change `FramesPageUnitTest` to wait for the frames (instead of sleep)
before asserting its state, also, wait for the expected state as the
state might not have been changed yet.

Part of #344 - Make functional tests more reliable


thc202:
Fixed the code format.

PR #423, opened 2019-02-21
Add basic telemetry WIP

psiinon:
This adds 3 hidden images that try (and fail;) to load bitly links (which 404):

- When the HUD initializes: https://bit.ly/owaspzap-hud-0-4+
- First page of the tutorial: https://bit.ly/owaspzap-hud-tutorial-start+
- Last page of the tutorial: https://bit.ly/owaspzap-hud-tutorial-end+

The above links are for the relevant stats, the actual URLs dont have the trailing plus.

Note that the initialize link has a hardcoded version in it - I think this is really useful but we'll need to work out if we're ok manually changing it each time or if we can automate it more effectively.

I realise there's a huge amount of telemetry we're _really_ like but this seemed to be to be a minimal useful initial set :)

Feedback from all appreciated!


psiinon:
Closing in favour of #431 which should provide a much more comprehensive solution.

PR #429, opened 2019-02-24
Chore/add docker compose

jaywon:
Added simple `docker-compose` file for easily running the project in a Docker container with minimal setup and command line interaction.

Could add explicit `docker network` to run on, not sure if anyone has an opinion on that or use the default that will be created.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+ZAP_API_KEY=
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@
 # JS
 node_modules/
 dist/
+
+# Docker/Docker Compose
+.env
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ Learn more:
 
 ## Using the HUD
 
-### Downloading
+### Usage
+
+#### From Source Code
 You can download ZAP enabled with the HUD from any of: 
 
 1. Download and run the latest [ZAP Weekly Release](https://github.com/zaproxy/zaproxy/wiki/Downloads#zap-weekly)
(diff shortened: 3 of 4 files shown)


thc202:
IMHO we should not repeat the same information in multiple places, either keep it here (probably better) or in the wiki, not both. Otherwise we have to maintain it in two places, in which case both will be out of sync/outdated very easily.

I'd suggest adding the files to a `docker` directory to avoid proliferation of Docker related files in the top level directory, also, one can add a more complete `README` there (with the information from the wiki) and just link to it from the main `README`.

dscrobonia:
I love this, @jaywon !

jaywon:
@thc202 I incorporated your feedback for moving the `docker` related files into it's own directory and did some copy adjustment to the main `README`. It feels a little strange going into a subdirectory to bring up the project but that's ok, I don't feel strongly about it. Usually this setup I would leave the `docker-compose` in the main directory and put all other `Dockerfile` and related assets into another directory and set the context appropriately but this is a bit different use case, so s/b G2G on that.

In regards to the documentation, I totally agree that info should not be duplicated but I also don't think it's intuitive or standard to have multiple `README`(s) in multiple places, so I didn't put a `README` in the `docker` directory for detailing instructions there. But have a few proposed solutions! 😄 

In general I think developers expect to get a few things out of a main `README` at a minimum:

- How do I install the project
- How do I run the project
- Basic configurations/usage details

Outside of that, it's standard expectation I think to go to a single wiki or branded documentation site for more descriptive information/details. 

I did start [this](https://jaywon.github.io/zap-hud-wiki/) and @mellaniesori and @dscrobonia were going to jump on a call this week to really get some traction around documentation with a proposed workflow if that's ok.

Ultimately, Github Wikis really break down for really "selling" a project and organizing information well as opposed to something like this. So....long story short, I'd like to propose documentation driving towards that. Right now it will exist in wiki and the main `README` but not for very long. Hope that wasn't too wordy! ✏️

PR #431, opened 2019-02-27
Feature/segment telemetry [WIP]

dscrobonia:
I need to edit the ZAP event bus to ignore analytics messages going to segment.io. Intercepted messages and history get blown up with analytics calls.

Working on getting the segment account bumped up a couple tiers for free, then I can invite y'all to the workspace.

There are also a few pieces of info that are included by default, that I'm trying to remove. Namely, userAgent and referer (target site). Once those two pieces of information are removed everything should be good to go!

### DIFF
--- a/src/main/java/org/zaproxy/zap/extension/hud/HudAPI.java
+++ b/src/main/java/org/zaproxy/zap/extension/hud/HudAPI.java
@@ -67,11 +67,11 @@ public class HudAPI extends ApiImplementor {
 
     // TODO shouldnt allow unsafe-inline styles - need to work out where they are being used
     protected static final String CSP_POLICY =
-            "default-src 'none'; script-src 'self'; connect-src https://zap wss://zap; frame-src 'self'; img-src 'self' data:; "
+            "default-src 'none'; script-src https://zap https://cdn.segment.com; connect-src https://zap wss://zap https://api.segment.io; frame-src 'self'; img-src 'self' data:; "
                     + "font-src 'self' data:; style-src 'self' 'unsafe-inline' ;";
 
     protected static final String CSP_POLICY_UNSAFE_EVAL =
-            "default-src 'none'; script-src 'self' 'unsafe-eval'; connect-src https://zap wss://zap; frame-src 'self'; img-src 'self' data:; "
+            "default-src 'none'; script-src https://zap https://cdn.segment.com 'unsafe-eval'; connect-src https://zap wss://zap https://api.segment.io; frame-src 'self'; img-src 'self' data:; "
                     + "font-src 'self' data:; style-src 'self' 'unsafe-inline' ;";
 
     private static final String PREFIX = "hud";
--- a/src/main/zapHomeFiles/hud/drawer.js
+++ b/src/main/zapHomeFiles/hud/drawer.js
@@ -250,6 +250,7 @@ Vue.component('tabs', {
             if (!this.isOpen) {
                 this.openDrawer();
                 selectedTab.badgeData = 0;
+			    analytics.track('ToolButtonClicked', {name: selectedTab.name}, { context: { ip: "0.0.0.0" }})
             }
             else {
                 if (selectedTab.isActive) {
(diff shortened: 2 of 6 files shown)


psiinon:
Loading the segment libs from https://cdn.segment.com/ seems to significantly slow down the HUD start up. Can we include them with the HUD? FYI The first request took 7.1 secs according to ZAP, and that was probably reflected in the load time increase.
Alternatively can we request them in thebackground without blocking?
Also what will happen if someone uses the HUD locally with no internet access - will the HUD still work? Will it take even longer to load?

psiinon:
Just had a look at some of the data thats being sent to Segment:

{"context":{"ip":"0.0.0.0","page":{"path":"//zapCallBackUrl/-1608052698053129056","referrer":"https://localhost:8080/bodgeit/product.jsp?typeid=4","search":"?name=panel.html&url=https://localhost:8080&orientation=right&frameId=rightPanel&tabId=4392046-38","title":"","url":"https://zap//zapCallBackUrl/-1608052698053129056?name=panel.html&url=https://localhost:8080&orientation=right&frameId=rightPanel&tabId=4392046-38"},"userAgent":"Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:64.0) Gecko/20100101 Firefox/64.0","library":{"name":"analytics.js","version":"3.7.2"},"campaign":{}},"integrations":{},"properties":{"id":"rightPanel"},"event":"FrameLoaded","anonymousId":"3ee25832-b805-4d36-b3df-2fb8ebf61bcf","timestamp":"2019-02-28T17:19:52.578Z","type":"track","writeKey":"gWdJpKHr0tdlE5inzWqL0RwdOkGfwF5u","userId":null,"sentAt":"2019-02-28T17:19:52.580Z","_metadata":{"bundled":["Segment.io"],"unbundled":[]},"messageId":"ajs-76414cf6ff5cf933d33bbf7a7aed720f"}

Theres no way we can collect the target URL - thats completely off limits.

dscrobonia:
> Theres no way we can collect the target URL - thats completely off limits
Agreed. That and the userAgent should not be collected. I included that in the PR description. Working to get those removed.

> Can we include them with the HUD?
Yes. Instead of requesting it from the CDN we can just download that file and host in ZAP ourselves. I also think we should do this for reliability, simplicity, and performance reasons.

> Also what will happen if someone uses the HUD locally with no internet access - will the HUD still work?
See above. Hosting it ourselves will solve this.

> Alternatively can we request them in thebackground without blocking?
It shouldn't be blocking, so anywhere that it is we should change that.

psiinon:
Re 'hiding' the telemetry requests, this can easily be achieved by adding the relevant regexes to the list of URLs excluded from the proxy.
We can add it via the HUD automatically but would need a way for the user to disable this if they want to see them. That could just be a description of what they need to do manually?

dscrobonia:
> We can add it via the HUD automatically but would need a way for the user to disable this if they want to see them. That could just be a description of what they need to do manually?

I think we should have a check box that does this automagically, and in the background. And I also think that if you check the "I accept to send telemetry data to ZAP" that this box get automatically checked for you.

As a user I don't want to have to add a regex myself. As a user I also don't even want to see the regex in the same config option I would set myself, because then I'm confused as to why there is a regex already set up that I haven't configured.

But I also think that from the perspective of a user who wants to know what is happening, there should be a simple check box in the options (probably the HUD options, unless we want a separate telemetry option which doesn't really make sense in ZAP cuz it would only apply to the HUD) that says "Show telemetry traffic in ZAP history" so that I can inspect what we're sending.

So to recap I think in the HUD options there should be two options:
1. "Enable the Heads Up Display to send telemetry." with a link to the wiki with more information about what that means.
2. "Hide telemetry traffic from the ZAP HTTP history."

When a user starts the HUD they will be presented with the initial pop up "Would you like to enable the HUD to gather simple metrics and telemetry to help us improve?". In that pop up a link to the same wiki (or other local doc) explaining what that data we're collecting, what we're using it for, and how to change the settings to see it. 

If the user selects yes, both options 1 & 2 will be selected for them.

psiinon:
Closing as this currently seems to be a very invasive approach, and as @dscrobonia no longer works at Segment we'd likely have more of a problem getting access t othe stats :)

PR #432, opened 2019-02-27
Add telemetry option

psiinon:
Note that this does not actually add the telemetry!

@dscrobonia this also doesnt add any user information about the telemetry - I'm assuming your PR will do that :)
You can check the telemetry option via the API but right now there are no events to say when the telemetry option is changed - will you need those? Can add them in a future PR if they will be needed.

### DIFF
--- a/src/main/java/org/zaproxy/zap/extension/hud/HudParam.java
+++ b/src/main/java/org/zaproxy/zap/extension/hud/HudParam.java
@@ -68,6 +68,7 @@ public class HudParam extends VersionedAbstractParam {
     private static final String PARAM_ENABLE_ON_DOMAIN_MSGS =
             PARAM_BASE_KEY + ".enableOnDomainMsgs";
     private static final String PARAM_UI_OPTION_PREFIX = PARAM_BASE_KEY + ".uiOption.";
+    private static final String PARAM_ENABLE_TELEMETRY = PARAM_BASE_KEY + ".enableTelemetry";
 
     /**
      * The version of the configurations. Used to keep track of configurations changes between
@@ -103,6 +104,8 @@ public class HudParam extends VersionedAbstractParam {
     private boolean showWelcomeScreen;
 
     private boolean enableOnDomainMsgs;
+
+    private boolean enableTelemetry;
 
     private List<String> tutorialTasks;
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/zaproxy/zap/extension/hud/HudParam.java
```diff
@@ -217,6 +229,7 @@ protected void parseImpl() {
         tutorialTasks = convert(getConfig().getList(PARAM_TUTORIAL_TASKS));
         showWelcomeScreen = getConfig().getBoolean(PARAM_SHOW_WELCOME_SCREEN, true);
         enableOnDomainMsgs = getConfig().getBoolean(PARAM_ENABLE_ON_DOMAIN_MSGS, true);
+        enableTelemetry = getConfig().getBoolean(PARAM_ENABLE_TELEMETRY, true);
```
thc202: Disable by default if in dev mode?

psiinon:
Closing in favour of #431 which should provide a much more comprehensive solution

psiinon:
Arg, closed wrong one :P

psiinon:
Finally fixed the one comment :P 
Should be ready for review.

psiinon:
Updated again :)

src/main/resources/org/zaproxy/zap/extension/hud/resources/Messages.properties
```diff
@@ -37,6 +37,7 @@ hud.optionspanel.button.baseDirectory	= Change
 hud.optionspanel.label.baseDirectory	= Base Directory:
 hud.optionspanel.label.enabledForDesktop = Enable when using the ZAP Desktop
 hud.optionspanel.label.enabledForDaemon = Enable when using ZAP in daemon mode
+hud.optionspanel.label.enableTelemetry = Enable anonymized telemetry
```
kingthorin, line 40: "anonymized" implies that it was identifiable and was rendered unidentified before submission, which if we take another small step implies that something could go wrong.... Would it be better to say "anonymous"? (Which to me implies that there's no identifying info to start with.)

psiinon: Oops, lost track of this :/
Now changed :)

Issue #373, opened 2019-01-29
Show websockets messages in new tab


psiinon:
Similar (and next) to the history tab and as per the desktop Websockets tab.
I'm thinking that this should only show websockets messages to the current site being displayed? Or is that an option we give to users?
This work can be phased:

- [x] New tab showing websockets messages
- [x] Select message to display and resend
- [x] Option to resend message
- [ ] Support break with websockets

This will require some changes to the ZAP websockets add-on

psiinon:
For info I've just got the websockets messages being shown in a new tab, although they could be prettier ;)

dscrobonia:
> I'm thinking that this should only show websockets messages to the current site being displayed? Or is that an option we give to users?

Whatever route we go with here lets keep it consistent to how we're treating http messages. I think my default both http and websocket histories should only show traffic for the site/tab that you're in, and then there should be an easy checkbox to display traffic across all tabs. I want to say this isn't a trivial fix, but as a user I think that's how I would want to experience it.

jaywon:
This is possibly a new way to think about "scopes" and how they're defined/experienced?

dscrobonia:
Ya I've actually been thinking about that quite a bit. In some tools we run them against the domain/url we're currently in and don't actually use the scope. And our notion of "scope" in the HUD hadn't previously supported multiple domains, urls, or regexes yet. I started making progress towards that in #350 by trying to treat scope as a list of urls. But lots more needed to be done.

Ideally things like history, websocket I think should the current tab only by default. But that there should be filters for "all traffic" and "in scope traffic"

PR #445, opened 2019-03-04
Display full websocket message

psiinon:
This is part 2 of #373 it displays the message but I havnt implemented resend yet.

### DIFF
--- a/src/main/zapHomeFiles/hud/display.html
+++ b/src/main/zapHomeFiles/hud/display.html
@@ -22,7 +22,8 @@
             <simple-menu-modal :show="isSimpleMenuModalShown" :title="simpleMenuModalTitle" @close="isSimpleMenuModalShown = false"></simple-menu-modal>
             <break-message-modal :show="isBreakMessageModalShown" :title="breakMessageModalTitle" @close="isBreakMessageModalShown = false"></break-message-modal>
             <history-message-modal :show="isHistoryMessageModalShown" :title="historyMessageModalTitle" @close="isHistoryMessageModalShown = false"></history-message-modal>
-            <site-tree-modal :show="isSiteTreeModalShown" :title="siteTreeModalTitle" @close="isSiteTreeModalShown = false"></http-message-modal>
+            <websocket-message-modal :show="isWebsocketMessageModalShown" :title="websocketMessageModalTitle" @close="isWebsocketMessageModalShown = false"></websocket-message-modal>
+            <site-tree-modal :show="isSiteTreeModalShown" :title="siteTreeModalTitle" @close="isSiteTreeModalShown = false"></site-tree-modal>
         </div>
         
         <!-- a base modal component -->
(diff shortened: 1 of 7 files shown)


psiinon:
Just noticed that incoming messages are still flagged as 'Outgoin' - will need to fix that.
I'll also change the table to use the 1i8n'd direction strings too.

psiinon:
Fixed those problems. Also fixed the problem where JSON payloads were being displayed as [Object]

thc202:
The side effect of `JSON.stringify` the payload is that it might change it, special chars are quoted and the content itself, if plain string.

psiinon:
This is a bug in the websockets add-on - I'll put a PR in to fix that.

psiinon:
Have fixed the JSON encoding bugs in https://github.com/zaproxy/zap-extensions/pull/1954
so undone the conversion thats no longer needed here.
Hopefully good to go now...

src/main/zapHomeFiles/hud/display.html
```diff
@@ -278,6 +279,37 @@
...
+                            <tr>
+                                <td> {{ $t('message.websockets_message_field_payload') }} </td>
+                                <td> </td>
+                            </tr>
+                        </tbody>
+                    </table>
+                    <textarea class="form-input hud-websocket-message" rows="10" spellcheck="false" v-model="payload" :readonly="true"></textarea>
+                </div>
+                <div slot="footer">
+                    <!-- button class="btn btn-primary" @click="replay"> {{ $t('message.websockets_replay') }} </button-->
```
Pamplemousse: Comment leftover.

src/main/zapHomeFiles/hud/display.js
```diff
@@ -438,6 +438,49 @@ Vue.component('history-message-modal', {
...
 
+Vue.component('websocket-message-modal', {
+	template: '#websocket-message-modal-template',
+	props: ['show', 'title'],
+	methods: {
+		close: function() {
+			this.$emit('close');
+		},
+		replay: function() {
+			// TODO implement
```
Pamplemousse: `TODO` leftover.

src/main/zapHomeFiles/hud/tools/websockets.js
```diff
@@ -45,6 +45,28 @@ var WebSockets = (function() {
...
+		if (!data) {
+			throw new Error('Could not load WebSocket message details')
+		}
+
+		return utils.messageFrame(tabId, "display", {action:"showWebSocketMessage", config:data})
+		.then(data => {
+			// Handle button choice
+			if (data.buttonSelected === "replay") {
+				utils.log (LOG_DEBUG, 'websockets.showWebSocketMessageDetails', 'replay', data);
+				// TODO implement
```
Pamplemousse: `TODO` leftover.

psiinon:
Updated to support websocket replay :)

psiinon:
@Pamplemousse those were all deliberately left in there .. and have now been emabled / implemented :D

psiinon:
Hum, should probably disable Replay if the payload is not text, as we dont really support that - will implement that...

psiinon:
Replay now disabled for non TEXT messages.
Looks like there no tutorial for websockets - my bad ;)
I'll add that in a future PR if thats ok? Maybe with break support ;)

thc202:
The dialogue just allows to replay the message once, one has to reopen the dialogue to replay again.

psiinon:
Good spot!
Simple option would be to close the dialog I guess...

psiinon:
Changed to close the dialog after sending. This keeps it consistent with the HTTP one.
We can always change it in the future ...

Issue #355, opened 2019-01-16
Add option to change the HUD font size(s)


psiinon:
Esp useful for the History tab, but could well be useful for the other tools as well.

PR #449, opened 2019-03-08
Minor cleaning in bottom-drawer

Pamplemousse:


jaywon:
Thank you for the recommended changes! I would like to request that these not be merged or discussed though. 

I'm not opposed to removing empty placeholder classes in the .css file at all, but semantically it's a better practice to give the elements meaning with unique ids and shared class names as they relate to their purpose within the application and keep the styling separate and flexible to change.

For example, if the widths change or go to a flex layout we would have to change the class names with them as opposed to rules within them. Or if we need to target certain columns they could be confused with other columns that share a similar width but functionally different meaning.

Does that make sense?

kingthorin:
> I would like to request that these not be merged or discussed though.

That’s a typo right? I can understand holding off on merging, but, only if everyone is gonna talk it out 👍

thc202:
That makes sense and it was my first thought, we lose flexibility but do we really need it now? (CSS is not my domain so...)

jaywon:
OMG I'm sorry, I can see how that can be misread. I meant i would recommend not merging this change **but** I'm always happy to have a discussion about it 😄 

We don't need the flexibility at this time but I always try to build for longer term flexibility if possible. Very happy to discuss!

Pamplemousse:
My point of view on that:
  * Flexibility is brought by modularity, ie, one can arrange classes to **compose** styles. Example: if one wants an html element to look like a clickable button and a tile, it would add `class="clickable-button tile"` to it.
  * Avoid duplication: if a bunch of classes have the same rules, then surely those rules can be extracted into an new class that can be "shared" across elements.
  * It seems to be a common pattern across css frameworks. For instance, [Spectre](https://picturepan2.github.io/spectre/index.html) uses classes to set the width of elements inside its grid system, see: https://picturepan2.github.io/spectre/layout/grid.html.

dscrobonia:
Oh man! Interesting discussion here.

I'm no CSS wizard, but for what its worth I agree with @Pamplemousse argument about creating composable styles. That's the paradigm I've gotten used to in Spectre for sure. Having CSS classes called "widthX" seems a little weird, but seems to make sense because if we needed to add more specific functionality to  `<th class="width-75">` we would just add that functionality to a new class `title` and then make it  `<th class="width-75 title">`. 

@jaywon 's point also makes sense that its a little overplanning at this point, considering if we want to tweak a column now we have to create a class and add it as opposed to just modifying an existing class.

I'm cool with these changes and sticking with the paradigm we use in other places created by Spectre. But ultimately either way is fine by me.

psiinon:
We could always have names like "column-short", "column-medium" ?

Pamplemousse:
Agree with @psiinon .

#### Here are reasons why I proposed as is in the PR:

  * There are several (arbitrary?) different fixed width affected to the columns: 75, 95, 125, 150 ;
  * Rather than been fixed, would dynamic width be "better" (responsive)?
  * If going for something like `column-short`, `column-medium`, why not use the spectre's classes `col-<1-12>`?

#### And that would mean is a bit more work / decision

I tried to make the width of columns to be dynamic using spectre ; But then, I ran through inconsistencies between the head and body of the tables, are they are in two separate `table` elements (kind of a code smell, but it allows the header to stay visible when scrolling ; relates to #451).

In the end, I tried to not go too far, as I am not a UX wizard, and thought that there is some decision making regarding the design and front-end code that needed to be made.

#### Did not think at all about the fact that it could be user custom...

As you said, it could be a greater change.

jaywon:
Great feedback all around! I agree with most points made and I think it's gotten to this point more as POC and now we're starting to really talk about some of the front end **philosophy** which is great, where up to this point it'd been more focused on POC.

Also the sticky header was a very key UX requirement more so than the CSS widths, and that's why the table's were split. It was the best way to achieve the sticky header without a full rewrite of the table layout at that time. The width style classes of the Spectre _might_ work on table columns but were really intended to be used on adjacent block level elements in a fluid layout, so it might be a challenge with table columns. Haven't tried but I know that's not their intended use case.

I think to fully move forward with resolution, it's really ditching the `<table>` layouts completely and moving to a grid (`<div>` based) layout that would offer more fluidity and give the ability to leverage the existing classes and responsiveness of the Spectre framework. 

Should we make that a ticket and focus on a more comprehensive layout "fix" with the requirements of:

* Grid based/responsive layout for fluidity and consistency of layouts/columns
* Use of pre-existing framework width definitions
* Removal of fixed width definitions 

As part of the larger discussion, font-size (#355) also applies to this higher level discussion around the UI of this section. It makes it a larger task but seems to be more in line with setting a good direction going forward now that we're all discussing it.

Pamplemousse:
Completely agree with @jaywon on that.
Btw, you seem to have much more experience than I do regarding front-end dev, so happy to follow your lead if you want to raise an issue about philosophy and style in this area.

Removed the troublesome commit from the PR.
Looking forward to see more rework / cleaning of the code!

dscrobonia:
I love all of this discussion! If there is a way we can start organizing design/UX principles I think that would be extremely powerful. Like you mentioned @jaywon we've been operating in PoC mode with a bunch of frameworks/hacks that @psiinon and I glued together. But I agree that we should start evolving out of that and start defining/refining our approach to different aspects of the project.

> I think to fully move forward with resolution, it's really ditching the <table> layouts completely and moving to a grid (<div> based) layout that would offer more fluidity and give the ability to leverage the existing classes and responsiveness of the Spectre framework.
> 
> Should we make that a ticket and focus on a more comprehensive layout "fix" with the requirements of:
> 
> Grid based/responsive layout for fluidity and consistency of layouts/columns
> Use of pre-existing framework width definitions
> Removal of fixed width definitions
> As part of the larger discussion, font-size (#355) also applies to this higher level discussion around the UI of this section. It makes it a larger task but seems to be more in line with setting a good direction going forward now that we're all discussing it.

This sounds like a great idea!

Pamplemousse:
Poke @psiinon: I removed the troublesome commit, anything else preventing the merge?

PR #452, opened 2019-03-11
Support break for websockets

### LINKED ISSUE #373: Show websockets messages in new tab

psiinon:
Fixes #373
Depends on https://github.com/zaproxy/zap-extensions/pull/1956

Need to add a websockets tutorial page.
Only problem is that it reports: [Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "payload"

Cant see why its failing for this and not the HTTP one - anyone have any ideas?


psiinon:
Updated to match the latest changes pushed to https://github.com/zaproxy/zap-extensions/pull/1956

psiinon:
Just added a WebSockets tutorial page, without a task for now.

psiinon:
Its worth noting that websockets break currently doesnt work in daemon mode. I'm looking at that, but its going to be a ZAP issue rather than a HUD one.

psiinon:
@dscrobonia any suggestions re the vuejs error messages?

dscrobonia:
Hey @psiinon When I run this I don't get that Vue.js error. Are you still getting it?

dscrobonia:
I also don't get get any websocket messages intercepted or any indication that websocket breaking is working though. :/ I'm assuming your changes to ZAP Desktop made it to the weekly build by now so that running ./gradlew runZap would be sufficient?

thc202:
Depends which weekly you have, you might need to get a newer one (i.e. remove the directory `build/testZapInstall`).

psiinon:
If other people arent getting the error should we merge this?

dscrobonia:
> Depends which weekly you have, you might need to get a newer one (i.e. remove the directory build/testZapInstall).

@thc202 do you think we should remove this directory when running the `./gradlew clean` command?

thc202:
It does already, it removes everything in the build directory. I suggested removing just that one to keep the other build files (which can be reused in following builds).

psiinon:
@dscrobonia is this working for you yet?
Be really good to get it in ZAP 2.8.0...

dscrobonia:
Yeah looks good to me! Let's get it in!

src/main/zapHomeFiles/hud/display.js
```diff
@@ -678,6 +757,8 @@ document.addEventListener("DOMContentLoaded", () => {
 			historyMessageModalTitle: I18n.t("history_http_message_title"),
 			isWebsocketMessageModalShown: false,
 			websocketMessageModalTitle: I18n.t("websockets_message_title"),
+			isBreakWebSocketMessageModalShown: false,
+			breakWebSocketMessageModalTitle: I18n.t("break_http_message_title"),
```
dscrobonia: Oh one small change here! This should be 
```
breakWebSocketMessageModalTitle: I18n.t("break_intercpet_ws_title"),
```

dscrobonia: Its showing the wrong title on intecept. after that, lets merge it!

PR #458, opened 2019-03-18
Tweaked instructions on readme

psiinon:
Feedback from a couple of people is that they assumed all 3 steps were necessary because they were numbered. Maybe this will make things clearer?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,16 +19,22 @@ Learn more:
 ## Using the HUD
 
 ### Downloading
-You can download ZAP enabled with the HUD from any of: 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,14 +21,20 @@ Learn more:
 ### Downloading
 You can download ZAP enabled with the HUD from any of:
```
thc202: Maybe replace `download` with `use`?

psiinon: Changed to "try out" as per irc conversation

PR #477, opened 2019-05-02
Add CONTRIBUTING.md file

thc202:
Mention that the changelog should be updated if the change being done is
relevant.


thc202:
For sure, a lot more can/need be added, in follow up PR(s). ;) I don't plan to add more in this one.

PR #489, opened 2019-05-15
HUD Tutorial Fix Typos

kingthorin:
Per emailed feedback.

### DIFF
--- a/src/main/zapHomeFiles/hudtutorial/en_GB/AlertNotifications.task.html
+++ b/src/main/zapHomeFiles/hudtutorial/en_GB/AlertNotifications.task.html
@@ -1,5 +1,5 @@
 	<h2>Task</h2>
-	Click the button then keep your eyes on the bottom right hand side of the page - 
+	Click the Submit button then keep your eyes on the bottom right hand side of the page - 
 	a series of alert notifications should be shown there.
 	<form action="AlertNotifications" method="post">
 	<input type="hidden" name="anticsrf" value="<!-- CSRF -->"/>
--- a/src/main/zapHomeFiles/hudtutorial/en_GB/PageAlerts.html
+++ b/src/main/zapHomeFiles/hudtutorial/en_GB/PageAlerts.html
@@ -11,15 +11,15 @@
 	which by default are shown on the left hand side. This allows you to just focus on the potential issues on
 	your current page without being distracted by issues on other pages. 
 	<p>
-	Alerts have an associated risk, and all alerts of with the same level of risk are shown together:
+	Alerts have an associated risk, and all alerts with the same level of risk are shown together:
 	<ul>
 	<li><img src="page-alerts-high.png"> High Level Page Alerts
-	<li><img src="page-alerts-low.png"> Low  Level Page Alerts
 	<li><img src="page-alerts-medium.png"> Medium Level Page Alerts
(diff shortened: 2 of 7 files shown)


src/main/zapHomeFiles/hudtutorial/en_GB/SiteAlerts.html
```diff
@@ -15,8 +15,8 @@ <h1>Site Alerts</H1>
 	The alerts are grouped by risk in the same way as Page Alerts:
 	<ul>
 	<li><img src="site-alerts-high.png"> High Level Site Alerts
-	<li><img src="site-alerts-low.png"> Low  Level Site Alerts
 	<li><img src="site-alerts-medium.png"> Medium Level Site Alerts
+	<li><img src="site-alerts-low.png"> Low Level Site Alerts
```
kingthorin: Should we remove all occurrences of "Site" here as with the Page list?

In both cases should the `li` have closing tags?

src/main/zapHomeFiles/hudtutorial/en_GB/AlertNotifications.task.html
```diff
@@ -1,5 +1,5 @@
 	<h2>Task</h2>
-	Click the button then keep your eyes on the bottom right hand side of the page - 
+	Click the submit button then keep your eyes on the bottom right hand side of the page -
```
thc202: Start with upper case? (to match the button label)

src/main/zapHomeFiles/hudtutorial/en_GB/PageAlerts.html
```diff
@@ -11,15 +11,15 @@ <h1>Page Alerts</H1>
 	which by default are shown on the left hand side. This allows you to just focus on the potential issues on
 	your current page without being distracted by issues on other pages. 
 	<p>
-	Alerts have an associated risk, and all alerts of with the same level of risk are shown together:
+	Alerts have an associated risk, and all alerts with the same level of risk are shown together:
 	<ul>
-	<li><img src="page-alerts-high.png"> High Level Page Alerts
```
thc202: I think they were as is to match the label of the tools.

src/main/zapHomeFiles/hudtutorial/en_GB/SiteAlerts.html
thc202: For consistency, if we don't mind the differences in the labels.

Fine either way.

src/main/zapHomeFiles/hudtutorial/en_GB/PageAlerts.html
kingthorin: I'm not set on it one way or another. I see the value in matching it, but I also see Dave's point that it's kind of redundant. 

@psiinon tie break it? :smirk:

psiinon: Should def match the UI, and that means I think we should have both "Page" and "Site" in all places. If we dont make those things clear people wont understand the difference...

kingthorin: Ok, I'll revert the "Page" removals then.

kingthorin:
Should be good to go now,.

thc202:
Thank you!

psiinon:
Nice!

PR #492, opened 2019-05-16
Rename HUD frames

### LINKED ISSUE #484: Rename HUD frame names
The current HUD frame names are too generic and could easily clash with existing ones on the target site, eg 'management', 'left-panel' etc
Just adding something like 'zap-hud-' to the start of all of them should do the trick...

psiinon:
Fixes #484 

### DIFF
--- a/src/test/java/org/zaproxy/zap/extension/hud/ui/uimap/HUD.java
+++ b/src/test/java/org/zaproxy/zap/extension/hud/ui/uimap/HUD.java
@@ -46,12 +46,25 @@ public class HUD {
         this.webdriver = webdriver;
     }
 
-    public static By LEFT_PANEL_BY_ID = By.id("left-panel");
-    public static By RIGHT_PANEL_BY_ID = By.id("right-panel");
-    public static By BOTTOM_PANEL_BY_ID = By.id("bottom-drawer");
-    public static By DISPLAY_PANEL_BY_ID = By.id("main-display");
+    public static String LEFT_PANEL_ID = "zap-hud-left-panel";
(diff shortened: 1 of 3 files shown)


psiinon:
The waitForHistoryTab() etc methods snuck in - I was working on those for the Replay reg tests - those parts work so I left them in ;)

src/test/java/org/zaproxy/zap/extension/hud/ui/uimap/HUD.java
```diff
@@ -46,12 +46,17 @@ public HUD(WebDriver webdriver) {
...
+    public static By RIGHT_PANEL_BY_ID = By.id("zap-hud-right-panel");
+    public static By BOTTOM_PANEL_BY_ID = By.id("zap-hud-bottom-drawer");
+    public static By DISPLAY_PANEL_BY_ID = By.id("zap-hud-main-display");
+
+    public static By HISTORY_TABLE_BY_ID = By.id("history-messages");
+
     public static By HUD_BUTTON_BY_CLASSNAME = By.className("hud-button");
+
     public static int LEFT_PANEL_MIN_BUTTONS = 8;
+    public static String HISTORY_TAB_LABEL = "History";
```
kingthorin: Should this have some sort of i18n treatment?

psiinon: This is just for the tests, so as long as we just run them in English (the default) then we should be fine.
Obviously it would be great to have tests for the i18n as well...

psiinon:
Missed the hardcoded management frame name in the BadSiteUnitTest - thanks @thc202

PR #499, opened 2019-05-27
Fix tests with Firefox >= 67

thc202:
Add preference to keep proxying localhost with newer Firefox versions.


thc202:
Latest Travis CI builds were failing because of this, e.g.:
https://travis-ci.com/zaproxy/zap-hud/jobs/202457295#L565

PR #504, opened 2019-06-01
fixed history scrolling

### LINKED ISSUE #354: Auto scroll to latest history record
When the History tab is expanded, or if you leave it expanded while you browse, then it always displays the first requests.
I think it would be better if it always displayed the latest requests, ie scrolled to the bottom of the set, as per the Desktop UI.

dscrobonia:
fixes #354


thc202:
In Firefox 67 the target page is also scrolled, works fine with Chrome 74 though.

The WebSocket tab also needs this fix.

dscrobonia:
Nice find @thc202 . I'll fix it for the websockets. And I did just noticed the firefox target scrolling as well which is strange. I'll see if I can solve that real quick.

PR #531, opened 2019-06-28
Replace hardcoded http urls on upgraded domains

### LINKED ISSUE #530: Direct AJAX calls not getting upgraded to https
As per https://groups.google.com/d/msg/zaproxy-users/byxt8JvbkZU/PNd6ASRBBAAJ

This could cause problems with lots of modern web apps, so we should investigate this asap.

psiinon:
Fixes #530 
Well, at least for simple cases, and they will hopefully be the most common


psiinon:
Added junit tests :)

PR #536, opened 2019-07-02
Added option to view the changelog

### LINKED ISSUE #535: Show changes in the HUD
I suspect that most ZAP users dont read the change log.
I think it would be really useful to show the HUD specific changes actually in the HUD.
Ideally we would subtly highlight these and then clear that indication when the user has looked at the changes.
Options:

1. Add a Changes page to the Tutorial
1. Add a new Changelog icon next to the gear icon at the bottom
1. Add a new Changelog item to the options
1. ??

Whichever one we choose I think the relevant icon should have a star or plus which is cleared when the user visits that page / tool - we'd need to keep track of what the user has seen so that it can be enabled again for the next version

psiinon:
Fixes #535 

OK, so it doesnt add any notifications yet, but thats still planned

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
  - Support for the Ajax Spider
+ - A config page option for showing the changelog [#535](https://github.com/zaproxy/zap-hud/issues/535)
 
 ### Fixed
  - Direct AJAX calls not getting upgraded to https [#530](https://github.com/zaproxy/zap-hud/issues/530)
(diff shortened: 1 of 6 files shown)


psiinon:
changelog-settings
changelog-details

psiinon:
@dscrobonia screenshots above - the changelog data is read from the add-on and therefore from https://github.com/zaproxy/zap-hud/blob/develop/CHANGELOG.md so its up to us to put a suitable level of detail in that (re https://github.com/zaproxy/zap-hud/issues/535#issuecomment-507936000)

psiinon:
You might be asking too much of me there :P 
In this case I think it would just be a nice to have - I cant see many people wanting to keep going back to the HUD options...

PR #548, opened 2019-07-11
Bugfix/scrolling entire page

dscrobonia:


thc202:
Seeing some problems with WebSockets tab:
1. Go to `https://www.websocket.org/echo.html`;
2. Select WebSockets tab;
3. Connect and send some messages until it starts to scroll (as expected);
4. Disconnect and connect again;
5. Send a message and notice that the scroll goes to the top.

src/main/zapHomeFiles/hud/drawer.js
```diff
@@ -201,26 +227,14 @@ Vue.component('websockets', {
 
         eventBus.$on('updateWebSockets', data => {
             this.messages = this.messages.concat(data.messages);
+            this.scrollToBottom()
```
thc202: This was not added to `updateMessages`?

thc202:
btw, worth updating the change log with the fix.

dscrobonia:
ah the messageid's on the websocket events aren't unique :( That's why this latest scrolling issue popped up @thc202. Good find!

dscrobonia:
Should work well now. 

One behavior with it that I don't like though is that if you were to scroll up to find a specific request, and new request comes in, it will jump you to the bottom. :/ but scrolling is much better now.

dscrobonia:
couldn't figure out how to detect scroll on the table. would be nice to match the behavior of the network tab


Feel free to merge if y'all like the behavior. its probably worth fixing the way it is currently

psiinon:
Hum, that could also be _really_ annoying - jumping to the bottom when you've explicitly scrolled to find something :/ 
Imagine you're looking at a modern web app that keeps making XHR requests - you'd never be able to find anything before it jumped to the bottom again.

Cant we use scroll events to detect scrolling? https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event 

I think we want:

- Normal case, new request comes in we scroll to the bottom
- User scolls up, new request doesnt scroll
- User scrolls to the bottom, back to normal case

The other option is something like a checkbox for auto scrolling to the bottom.

thc202:
Yeah, that's the behaviour we have in desktop UI, if the scroll it's at the bottom it keeps scrolling on new entries otherwise it doesn't scroll (the auto-scroll can also be disabled).

dscrobonia:
Yeah that's definitely the behavior we want.  I spent a few hours trying to figure that out yesterday and  wasn't able to figure out where to tap for the `onscroll` event. I couldn't find any js hooks or DOM events to see that the user had scrolled. :/

PR #554, opened 2019-07-18
Flag HUD changes

### LINKED ISSUE #553: Flag changes in the HUD
Follow up from #535
This issue is for flagging the changes in the HUD UI.
I've tried little plus and star icons but unfortunately I dont think they are very noticeable :(
gear-plus gear-star
I'm now thinking that we need a much bigger notification indicator. Will play around and post here...

psiinon:
Fixes #553 

Screenshots are in the bug.
The exclamation icon on the gear icon only gets cleared when you load a new page - I tried to fix that with events but failed ;) I dont think its a biggie but feel free to disagree.
To test different cases just reset the HUD version to 1 in the config.xml file, eg `<hud version="2">` then restart ZAP.
This changes means we will have to update this version for each new HUD version, and remember to flag all tutorial pages that are changed...


psiinon:
Fixed lgtm issue

src/main/java/org/zaproxy/zap/extension/hud/HudParam.java
```diff
@@ -64,8 +71,10 @@
      * releases, if updates are needed.
      *
      * <p>It only needs to be updated for configurations changes (not releases of the add-on).
+     * However for the HUD we do use t to flag new features, so it will typically be updated for
```
kingthorin: It?

src/main/java/org/zaproxy/zap/extension/hud/HudParam.java
```diff
@@ -328,6 +376,16 @@ public void setUiOption(String key, String value) {
...
     }
 
+    public void clearNewChangelog() {
+        this.newChangelog = false;
+        getConfig().setProperty(PARAM_NEW_CHANGELOG, newChangelog);
+        try {
+            this.getConfig().save();
+        } catch (ConfigurationException e) {
+            log.error(e.getMessage(), e);
+        }
```
kingthorin: Extract a saveConfig method to use in these clear and update functions?

psiinon: too obvious ;)

psiinon:
Hopefully fixed all of the comments :)

src/main/java/org/zaproxy/zap/extension/hud/HudParam.java
psiinon: yep

thc202:
Thanks!

src/main/zapHomeFiles/hud/display.js
```diff
@@ -339,6 +339,40 @@ Vue.component('simple-menu-modal', {
...
 
+Vue.component('adv-menu-modal', {
+	template: '#adv-menu-modal-template',
+	props: ['show', 'title'],
+	methods: {
+		close: function() {
+			this.$emit('close');
+		},
+		itemSelect: function(itemId) {
+			this.port.postMessage({'action': 'itemSelected', 'id': itemId});
```
dscrobonia: you'll need: 			

```
app.keepShowing = true;
app.isAdvMenuModalShown = false;
```
Inserted here

dscrobonia: This will fix the issue I noticed

dscrobonia:
Great work on this one @psiinon !

src/main/zapHomeFiles/hud/display.js
psiinon: Weird, I couldnt reproduce the issue, either with Firefox or chrome.
I've added the lines you suggested as they dont seem to cause any problems here, but it would be good to know why you hit this problem while no one else did...

psiinon:
I'm afraid the gold star on the gear icon is all but invisible - see #553 
But I do understand your concerns re the red icon.
How about a green one?
exclamation-green

psiinon:
So @thc202 has pointed out that the green icon wont work on the tutorial index button :/
I think we should stick with the red one.

psiinon:
Its not like we are pushing out really frequent changes, and it means that people should notice them.

PR #560, opened 2019-07-24
Fixed tutorial bug

psiinon:
Fixed bug which meant the HUD became completely unresponsive when the tutorial was launched.
Deliberately being fixed in master as we're in the process of generating the release...


kingthorin:
Don’t forget to get it back into develop too :zap:

PR #561, opened 2019-07-24
Fix API JSON parsing

psiinon:
Another last min fix - this should fix the problems seen when resetting the HUD configs.
These were causing misc issues, all logged to the console...


psiinon:
@thc202 has pointer out that now when you press the green + buttons after a reset all tools are shown, rather than just those that are not currently used :/

thc202:
Merging, that's not caused by these changes (happens with older versions).

thc202:
For the record the above issue is being tracked in #562.

PR #571, opened 2019-08-02
Access files by path instead of query

psiinon:
I've not updated the changelog as this is an 'internal' change to support #459
URLs used by the HUD to request files and images will now have formats like:

- `https://zap//zapCallBackUrl/<random number>/file/panel.html`
- `https://zap//zapCallBackUrl/<random number>/image/target-grey.png`

### DIFF
--- a/src/main/java/org/zaproxy/zap/extension/hud/HudAPI.java
+++ b/src/main/java/org/zaproxy/zap/extension/hud/HudAPI.java
@@ -185,22 +185,22 @@ public class HudAPI extends ApiImplementor {
     public String handleCallBack(HttpMessage msg) throws ApiException {
         // Just used to handle files which need to be on the target domain
         try {
-            String query = msg.getRequestHeader().getURI().getPathQuery();
-            logger.debug("callback query = " + query);
-            if (query != null) {
-                if (query.indexOf("zapfile=") > 0) {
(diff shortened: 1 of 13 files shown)


src/main/java/org/zaproxy/zap/extension/hud/HudAPI.java
```diff
@@ -185,22 +185,22 @@ public boolean allowUnsafeEval() {
...
+
+            logger.debug("callback fileName = " + fileName);
+            if (fileName != null) {
+                if (DOMAIN_FILE_WHITELIST.contains(fileName)) {
+                    msg.setResponseBody(
+                            this.getFile(msg, ExtensionHUD.TARGET_DIRECTORY + "/" + fileName));
+                    // Currently only javascript files are returned
+                    msg.setResponseHeader(
+                            API.getDefaultResponseHeader(
+                                    "application/javascript; charset=UTF-8",
```
jaywon: Should this have a little more evaluation on file type for setting the header since we'll be serving CSS and image files as well?

thc202: Those will be served by `HudFileProxy` which does that.

dscrobonia:
So exciting!!!

PR #575, opened 2019-08-07
Correct Content-Length of upgraded responses

thc202:
Use the length of the response body instead of request.

---
Fixes the issue reported in OWASP Slack channel when accessing Juice Shop.


kingthorin:
Should we be releasing this? Won’t it be impacting people’s use of the addon?

thc202:
> Should we be releasing this?

Yes, that's why I requested a review from you two.

> Won’t it be impacting people’s use of the addon?

It might, depends on the contents of the files (afaik we only have one report so far).

psiinon:
We know it affects someone, and have no idea how many others are affected who havnt reported it.
I think we should release the fix asap - first impressions are key and if the HUD just doesnt work then people might just leave it...

PR #609, opened 2019-09-26
Tidy up Java code

thc202:
Address some SonarCloud issues and other tidy ups:
 - Add missing `@Override` annotations;
 - Remove redundant type arguments;
 - Replace event listeners, `Function`, and `Thread` override with
 lambda usage;
 - Use `String` methods with `char` (e.g. `indexOf`, `lastIndexOf`);
 - Remove unnecessary array instantiation (varargs).
 - Remove declaration of runtime exceptions and of exceptions not
 actually thrown.
 - Remove unnecessary extension constructor (never used/called).
 - Address tasks post 2.8.0 update.


psiinon:
Thanks - this was on my todo list but you beat me to it :D

Issue #608, opened 2019-09-26
Fix lint warnings


psiinon:
The following rules were set to "warn" so that we could focus on more important issues, but these should all be fixed as well:

- [ ] no-negated-condition
- [ ] no-unused-vars
- [ ] prefer-destructuring
- [ ] unicorn/no-for-loop
- [ ] unicorn/prefer-includes
- [ ] unicorn/prefer-string-slice
- [ ] unicorn/filename-case
- [ ] unicorn/prefer-query-selector

The overrides are defined in https://github.com/zaproxy/zap-hud/blob/develop/package.json#L54-L63 (the lines may change in time)

njmulsqb:
These are the pending no-warning-comments in HUD.

1. // TODO change to try loading from localstorage (src\main\zapHomeFiles\hud\tools\commonAlerts.js)
2. // Todo: change this to 'continue' and figure out / fix stopBreaking (src\main\zapHomeFiles\hud\tools\break.js)
3. // FIXME: remove after #620 (src\main\zapHomeFiles\hud\display.js)
4. 	// Todo: is this too hacky? (src\main\zapHomeFiles\hud\target\inject.js)
5. // TODO showComments section - put this code in a separate file and inject ? (src\main\zapHomeFiles\hud\target\inject.js)
6. // TODO showEnable section - put this code in a separate file and inject ? (src\main\zapHomeFiles\hud\target\inject.js)

Give me the go-ahead to remove them and then we can merge this PR

psiinon:
Assigned 😁

PR #620, opened 2019-10-08
introduce eslint for DOM XSS linting

mozfreddyb:
For this to be immediately useful, one needs to:
- [ ] fix existing violations (below) #621
- [x] ~in package.json, add "script" to run eslint on all relevant source files~
- [x] integrate script into CI

Existing violations:
```
zap-hud/src/main/zapHomeFiles/hud/display.js
  1116:5  error  Unsafe assignment to innerHTML  no-unsanitized/property
zap-hud/src/main/zapHomeFiles/hud/libraries/alertify.js
  1:2257  error  Unsafe assignment to innerHTML  no-unsanitized/property
  1:2897  error  Unsafe assignment to innerHTML  no-unsanitized/property
zap-hud/src/main/zapHomeFiles/hud/target/inject.js
  454:3  error  Unsafe assignment to innerHTML  no-unsanitized/property
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,8 +51,25 @@
       "src/test/js/**",
       "src/webpack/**"
     ],
+    "plugins": ["no-unsanitized"],
     "rules": {
       "no-negated-condition": "warn",
+      "no-unsanitized/method": ["error",
+        {
+          "escape": {
+            "methods": [
+              "DOMPurify.sanitize"
+            ]
+          }
+        }],
+      "no-unsanitized/property": ["error",
+        {
+          "escape": {
+            "methods": [
+              "DOMPurify.sanitize"
+            ]
+          }
+        }],
       "no-unused-vars": "warn",
       "prefer-destructuring": "warn",
       "unicorn/no-for-loop": "warn",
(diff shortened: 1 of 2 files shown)


thc202:
Can't this be configured through `xo`?

psiinon:
For info have just raised #621re removing the use of innerHtml

mozfreddyb:
xo seems to use eslint underneath, so theoretically yes. I haven't ever used or looked at xo before though.

thc202:
OK, `xo` is already running by default (locally and CI) and (I think) it includes all relevant source files, so it would just need the plugin(s).

mozfreddyb:
OK. Waiting for #621 then :-)

psiinon:
Can we whitelist the remaining 2 uses of innerHtml so that we can merge this PR?
If so we should add them to #608 so they dont get forgotten.
I think we should ignore everything under zap-hud/src/main/zapHomeFiles/hud/libraries/ as they are external libs.

psiinon:
Fine for display.js but I dont want to apply DOMPurify to the alertify.js library - no idea what impact that could have.

mozfreddyb:
it seems alertify.js is already opted out of the linter. probably because it's minified?

psiinon:
And because its an external lib :)

PR #635, opened 2019-11-06
Add tweet link to the completion screen

psiinon:
Uses a free account with https://clicktotweet.com/ associated with the @zaproxy twitter account.
Users wont be logged into twitter as its a new profile, so they will see a screen like this:

Screenshot_2019-11-06 Post a Tweet on Twitter


kingthorin:
Do we need to be clear about the fact that they don't have to login somehow?

Also shall we use this as an opportunity to start using `#YouDontKnowZap`?

psiinon:
Yeah, so they will need to login to tweet :)
Re #YouDontKnowZap  - might it be better to focus on "ZAP in 10" for now, as thats nearly ready to go?

PR #638, opened 2019-11-22
Handle sites that fail when 'only in scope' switched on

### LINKED ISSUE #316: Sites upgraded to https fail if 'only in scope' switched on
Start using the HUD with an HTTP based site, but dont add it to the scope. The HUD will upgrade it to HTTPS.
Turn on the 'Enable the HUD only for URLs that are in scope' option.
The site will then fail with "Unrecognized SSL message, plaintext connection?" errors.
This is pretty confusing, especially for newcomers.
(Thanks mum: https://twitter.com/psiinon/status/1077890840142450689 ;)

psiinon:
Fixes #316

To test:

1. Launch browser with HUD
1. View site, HUD should be on
1. In ZAP set HUD option to only enable it if in scope
1. Follow an app link in the browser

Previously this would have failed, it should now work.

### DIFF
--- a/src/main/java/org/zaproxy/zap/extension/hud/HttpUpgradeProxyListener.java
+++ b/src/main/java/org/zaproxy/zap/extension/hud/HttpUpgradeProxyListener.java
@@ -30,6 +30,7 @@ import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.core.proxy.OverrideMessageProxyListener;
 import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpResponseHeader;
 import org.zaproxy.zap.ZAP;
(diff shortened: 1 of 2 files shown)


src/main/java/org/zaproxy/zap/extension/hud/HttpUpgradeProxyListener.java
```diff
@@ -55,32 +56,40 @@ public int getArrangeableListenerOrder() {
...
     public boolean onHttpRequestSend(HttpMessage msg) {
         if (this.extHud.isHudEnabled()) {
-            if (this.extHud.getHudParam().isInScopeOnly() && !msg.isInScope()) {
-                return false;
-            }
             try {
+                URI uri = msg.getRequestHeader().getURI();
+                if (this.extHud.getHudParam().isInScopeOnly() && !msg.isInScope()) {
+                    if (this.extHud.isUpgradedHttpsDomain(uri)) {
+                        // 302 to the https version..
```
kingthorin: Non-s

psiinon: Doh! Fixed

PR #641, opened 2019-11-26
Bugfix/Alert i18n bug #235

### LINKED ISSUE #235: Alert i18n bug
The alert internationalisation isnt quite right.
In [UIMessages.properties](https://github.com/psiinon/zap-hud/blob/master/src/other/resources/UIMessages/UIMessages.properties) set:
`alerts_risk_info = Info`
(if its not already).
Run the HUD and click on the Info alerts tool - the alerts dialog will be shown but the Info panel will not be selected. Thats because the code is using the string Informational, probably from here:
* https://github.com/psiinon/zap-hud/blob/master/src/main/zapHomeFiles/hud/tools/pageAlertsInformational.js#L18
* https://github.com/psiinon/zap-hud/blob/master/src/main/zapHomeFiles/hud/tools/siteAlertsInformational.js#L18

It should be using the i18n'd messages instead.
It might not be a trivial fix, the hardcoded string is probably used to access the alerts from the ZAP API, so we'll need to make sure the right string is used at the right time.
My guess is that the alert selection will fail for other languages, eg French.

Rshahatit:
Worked with David on this today. Figured out it was pretty much what Simon was saying in the issue description. The Alert_Risk wasn't getting set to the translated word, but we couldn't change it everywhere because it was also being used to make api calls. So we opted to create another constant with the correct translated function call and implement that only in the showAlerts function. 

Fix #235.

I tried running the tests with npm run test but it was getting an error on 
```
ERROR in ./src/main/zapHomeFiles/hud/serviceworker.js 4:3
Module parse failed: Unexpected token (4:3)
File was processed with these loaders:
 * ./src/webpack/loaders/zap-loader.js
You may need an additional loader to handle the result of these loaders.
| const ZAP_HUD_FILES = 'https://zap';
| const toolScripts = [
> 	''http://zap/to/some/tool'' <<<these are two single quotes not double quotes>>>
| ];
```
I went to zap-loader.js and found that there are escaped single quotes and it said they are there to make the js linter happy but when i removed them the test completed :) I didnt push that change here but just wanted to mention it. 
`source = source.replace('<<ZAP_HUD_TOOLS>>', '\'http://zap/to/some/tool\'');`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
+### Fixed
+- Alerts open with correct active tab in other supported languages besides English (ie French). [#235](https://github.com/zaproxy/zap-hud/issues/235)
 
 ## [0.8.0] - 2019-11-25
 
--- a/src/main/zapHomeFiles/hud/tools/pageAlertsHigh.js
+++ b/src/main/zapHomeFiles/hud/tools/pageAlertsHigh.js
@@ -16,6 +16,7 @@ const PageAlertsHigh = (function () {
 	ICONS.PA = 'page-alerts-high.png';
 	const ALERT_TYPE = 'page-alerts';
 	const ALERT_RISK = 'High';
+	const ALERT_RISK_LABEL = I18n.t('alerts_risk_high');
 
 	// Todo: change this to a util function that reads in a config file (json/xml)
 	function initializeStorage() {
@@ -36,7 +37,7 @@ const PageAlertsHigh = (function () {
 	}
 
 	function showAlerts(tabId, url) {
-		alertUtils.showPageAlerts(tabId, DIALOG, url, ALERT_RISK);
+		alertUtils.showPageAlerts(tabId, DIALOG, url, ALERT_RISK_LABEL);
 	}
 
 	function showOptions(tabId) {
--- a/src/main/zapHomeFiles/hud/tools/pageAlertsInformational.js
+++ b/src/main/zapHomeFiles/hud/tools/pageAlertsInformational.js
@@ -16,6 +16,7 @@ const PageAlertsInformational = (function () {
 	ICONS.PA = 'page-alerts-informational.png';
 	const ALERT_TYPE = 'page-alerts';
 	const ALERT_RISK = 'Informational';
+	const ALERT_RISK_LABEL = I18n.t('alerts_risk_info');
 
 	// Todo: change this to a util function that reads in a config file (json/xml)
 	function initializeStorage() {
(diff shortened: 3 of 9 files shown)


src/main/zapHomeFiles/hud/tools/pageAlertsHigh.js
```diff
@@ -16,6 +16,7 @@ const PageAlertsHigh = (function () {
 	ICONS.PA = 'page-alerts-high.png';
 	const ALERT_TYPE = 'page-alerts';
 	const ALERT_RISK = 'High';
+	const ALERT_RISK_TEXT = I18n.t('alerts_risk_high');
```
thc202: Maybe "label" instead of "text"?

thc202:
There's a lint issue in `pageAlertsLow.js`.

Rshahatit:
> There's a lint issue in `pageAlertsLow.js`.
I see a lint issue in every file actually because of the todo comments and the file name not being in kebab case. Should I make those changes?

thc202:
Sorry, was referring just to the error:
```
✖    20:1   Trailing spaces not allowed.            no-trailing-spaces
```
(The warnings are  ok to leave for now, there's already an issue to address them.)

thc202:
btw, worth mentioning the fix in the changelog.

src/main/zapHomeFiles/hud/tools/pageAlertsHigh.js
Rshahatit: i made this change! thank you for the suggestion.

Rshahatit:
> mentioning

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
+### Fixed
+- Alerts open with correct active tab in other supported languages besides English (ie French). [#235]
```
thc202: Could you add the link to the issue as well?

thc202:
The author/committer email looks unusual, was that set on purpose? https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address

CHANGELOG.md
Rshahatit: I added the link!

Rshahatit:
> The author/committer email looks unusual, was that set on purpose? https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address

Thanks for making me aware of that. I probably did that a long time ago and forgot about it. I updated it so hopefully its showing my proper email address now.

CHANGELOG.md
thc202: Thanks.

thc202:
You'd have to amend the commits (also with `--reset-author`) to update the email in those.

Rshahatit:
Had a little trouble with git in changing the author, I ended up rewriting those commits with the correct email. Not sure if I did it the correct way but I made sure to merge with the existing changes so hopefully its okay.

thc202:
I'd suggest squash all of them, to remove the old commits (or just drop those).

dscrobonia:
Thanks for getting all the linting and changelog corrections lined up @thc202 ! Anything else @Rshahatit needs to do in order to merge this commit?

thc202:
No, looks good (I didn't notice nor get notification of the latest push).

thc202:
Thanks!

PR #666, opened 2020-01-17
Prepare for next release

psiinon:


kingthorin:
There's something wrong here, one file shows 0.9.0 the other 0.10.0

psiinon:
I think thats normal isnt it? 9 is the current version, 10 will be the next one (but I could be wrong) @thc202 ?

thc202:
Right.

kingthorin:
Sorry was reading it as added content not just shifted...

psiinon:
But it sounds like I've messed up one of the other commits so I'll look at this again

psiinon:
Hopefully fixed now thanks again to @thc202 :)

PR #671, opened 2020-01-22
Only in scope icon

### LINKED ISSUE #670: Only in scope usability issue
If the "Enable the HUD only for URLs that are in scope" option is selected then it can appear that the HUD is broken: https://github.com/zaproxy/zaproxy/issues/5813
We've hit this a couple of other times as well.

Possible solutions:

- Only enable this mode if something is in scope (I have a feeling this was how it was originally implemented??)
- Change the icon, eg to include a target icon
- Never persist the option so the user has to always turn it on

Should this actually be controlled via the HUD? Eg turning it on and off?

psiinon:
Fixes #670 
In conjunction with https://github.com/zaproxy/zap-extensions/pull/2279


psiinon:
hud-scope-warning

Screenshot with those 2 changes applied

Issue #701, opened 2020-03-12
HUD broken with Firefox 74.0


psiinon:
HUD is not shown on the first page visited (in this case a local bodgeit).
It appears when the next page is viewed, but HUD tools fail - nothing happens when you click on them.
Lots of Vue errors in the console: "SyntaxError: An invalid or illegal string was specified".
Initially reported vi the User Group: https://groups.google.com/d/msg/zaproxy-users/m-qH7lSao6E/KI2Q460kAgAJ
Chrome 80 appears to still work.

psiinon:
It looks like the referrer value is not being set correctly, which could be due to either https://bugzilla.mozilla.org/show_bug.cgi?id=1601496 or https://bugzilla.mozilla.org/show_bug.cgi?id=1601743 :/

psiinon:
Update - various of the HUD files use document.referrer:

- drawer.js
- display.js
- panel.js
- management.js

My current plan is to replace these with a constant and to inject the referrer taken from the header via ZAP.
Initial attempt got some of the HUD working (alerts) but not all of it - getting errors that the target origin provided doesnt match the recipient windows origin, so looks like I may have replaced too much?

psiinon:
Interesting - I just tried a very simple test in FX 74 and the document.referrer element _is_ correct when linking between pages on one site. It also seems to work on https://whatismyreferer.com/
That implies theres something else going on I dont yet understand...

psiinon:
I've been reminded that FX 74 introduced Enhanced Tracking Protection, which removed the referrer when tracking cookies are detected. This sounds possible, but I've just tried disabling ETP with the HUD but it still breaks :(

dscrobonia:
Without having looked to deep into it, the idea is that one of our hacks of passing some info to the HUD from ZAP via the referer header is failing because FF is removing referer info?

psiinon:
Yep, I think thats it :/

dscrobonia:
I wonder if we can just pass it through via a GET or POST parameter and carve it out in the ServiceWorker's proxy capability...

tavdog:
In the meantime, is there a last known version of FF that works ?  Nevermind, I just installed version 73 and the HUD is working.

thc202:
The HUD stopped working when this change was introduced: https://bugzilla.mozilla.org/show_bug.cgi?id=1583700
[...]

PR #735, opened 2020-04-22
Unpin Firefox version in Travis CI

thc202:
Change Travis CI configuration to use latest as before.
Add the preference to restore the previous behaviour which makes the HUD
work again.

Ref #701.


thc202:
Forgot the tests are not using browsers launched by ZAP, added the required preference to the tests.

PR #791, opened 2020-08-06
Scale the HUD button

psiinon:
Dont think this is worth mentioning in the changelog, but can if anyone disagrees :)

To test just change the display font size to something significantly different and restart ZAP.
Without this fix the HUD button is not correctly resized.


kingthorin:
Travis seems to be failing due to broken dependencies?

thc202:
Yes, NPM is having problems (being fixed).

PR #836, opened 2020-11-18
Just warn on lint unicorn/prefer-ternary

psiinon:
To fix https://travis-ci.com/github/zaproxy/zap-hud/jobs/442226410
Not sure why its failing on that PR and not others, but only really want to warn on this anyway..

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,8 @@
       "unicorn/prefer-includes": "warn",
       "unicorn/prefer-string-slice": "warn",
       "unicorn/filename-case": "warn",
-      "unicorn/prefer-query-selector": "warn"
+      "unicorn/prefer-query-selector": "warn",
+      "unicorn/prefer-ternary": "warn"
     }
   },
   "repository": {


kingthorin (approved):

thc202:
This will have to be added in the other PR, it's a new rule.

Why "suppress" instead of changing the code? It would be more clear that's being initialised the variable.

psiinon:
TBH I think:
```
		if (url.indexOf('://') > -1) {
			hostname = url.split('/')[2];
		} else {
			hostname = url.split('/')[0];
		}
```
is easier to read than 
```
		let hostname = url.indexOf('://') > -1 ? url.split('/')[2] : hostname = url.split('/')[0];
```
but happy to change that if I'm in the minoroty...

thc202:
I prefer the later, but don't mind keeping what we have now. If we don't want the rule it's probably better to just disable than warn.

thc202:
Closing this one as the change needs to be applied in #840. Could you apply the change there?

PR #841, opened 2020-11-21
Support JDK 14

jveldhuijzen:
Upgrade the project to handle JDK14


jveldhuijzen:
Hi guys,
I wanted to check the repo out and see if I could contribute. I was installing java (JDK 14) and tried to run `gradlew runZap` but that kept failing:

> Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7

Solution: upgrade gradle:
https://github.com/gradle/gradle/issues/10248

So I've updated gradle and spotless and fixed some issue that came with the upgrade.

When you guys do not want to upgrade, please let me know what JDK should be used (and also add that to the [Wiki](https://github.com/zaproxy/zap-hud/wiki/Developing-with-the-HUD))

psiinon:
While we appreciate that its always good to use the most recent versions of java we also know that many ZAP users do not want to migrate so quickly. We dont get much fedback but based on this years survey ~ 30% are on java 8 and more than 50% on 11, 12 or 13. That will have changed since some people have answered but we still need to be very careful before forcing users to upgrade.
Obviously we can generate jars for older versions but we do need to make sure that this works consistantly.

thc202:
Closing per above comment. We are also targeting the same Gradle version throughout the ZAP repos, will be updating them all after the main ZAP release.

jveldhuijzen:
I agree that closing this was the right action.
Maybe add some documentation in the development wiki to describe the requirements for the dev env?

thc202:
Wiki [dev page](https://github.com/zaproxy/zap-hud/wiki/Developing-with-the-HUD) updated with recommended Java versions.

PR #843, opened 2020-11-24
Lint javascript for license header and add licence to js files

### LINKED ISSUE #1: License Header
I assume our standard license blurb should be added to all java files?
https://github.com/psiinon/zap-hud/blob/master/src/org/zaproxy/zap/extension/hud/HudAPI.java

What about all the JS files?

jveldhuijzen:
Adds license header files to javascript and lints the javascript files.
Fixes #1


src/main/zapHomeFiles/hud/display.js
```diff
@@ -1,3 +1,23 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright $YEAR The ZAP Development Team
```
thc202, line 6: It has the year as token.

src/main/zapHomeFiles/hud/display.js
```diff
@@ -1125,8 +1145,6 @@ navigator.serviceWorker.addEventListener('message', event => {
 
 			channel.port1.addEventListener('message', event => {
 				// Open window and inject the HTML report
-				// FIXME: remove after #620
-				// eslint-disable-next-line no-unsanitized/property
```
thc202, line 1129: This should be reverted.

thc202:
Changed to WIP per above comments.

thc202:
@jveldhuijzen do you still plan to work on this?

jveldhuijzen:
Well.. it's been busy. Where I left this off was that the lib used to to the linting (and fixing the header) wasn't able to do a replacement of the `$year` variable. 
I was busy with extending that lib (eslint-plugin-license-header) and wanted to setup a PR for them but all the unit tests for that project kept failing on my W10 machine, and I was planning on trying this out in a linux vm to see what would happen there.

So yeah, I'll try that project out in a linux vm to see where it leads me with the unit testing. 
I'll report back soon.

thc202:
Thank you.

jveldhuijzen:
As you may've wondered, I'm not getting around fixing this the right way. So I'll leave this PR for someone else to continue upon or do you guys want to close this?
Running the test for that other project does work on a linux machine, not a Windows machine.

nothingismagick:
Its funny, I've been looking for something like this for JS (after building it in rust over at https://github.com/tauri-apps/tauri ) - I have a couple cycles free this week and can pick this one up if you like @thc202

thc202:
That sounds good to me, I guess you would have to create a new PR (as you'll not be able to change this?). We can leave it open in the meantime though.

nothingismagick:
Would you prefer a proper SPDX approach? If so, who is the exact copywrite holder?

thc202:
I don't think it needs to be but if it addresses the issue. We go with "The ZAP Development Team".

PR #893, opened 2021-02-24
Accessibility fixes

zersiax:
This PR adds some ARIA properties to the current UI, making it easier for screenreaders to infer what particular components are meant to be doing. In particular:

- Some unlabeled controls are now labeled
- - Modals will behave more like native modals where screenreaders are concerned, which also adds a quick way to navigate to them from the frames, as currently modals that open don't redirect screenreader focus
- tablists and tabs now have the appropriate semantic roles.

While ARIA is generally discouraged in favor of using semantic, native HTML, the current incarnation of the HUD would require quite the significant rewrite to make depending on semantic HTML viable.
While I am reasonably certain this won't break anything, I was unable to get the project to build to verify this. Probably a good idea to quickly pull down the branch to see if I didn't break any Vue magic unintentionally. ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix typo in tutorial.
-
+- Added initial accessibility fixes for modals and tabs
 ## [0.12.0] - 2020-10-15
 ### Fixed
 - Problems with Firefox 81 due to referer header not being set cross domain. [#815](https://github.com/zaproxy/zap-hud/issues/815)
--- a/src/main/zapHomeFiles/hud/display.html
+++ b/src/main/zapHomeFiles/hud/display.html
@@ -32,13 +32,13 @@
         <!-- a base modal component -->
         <template id="modal-template">
             <transition name="modal" @after-leave="afterLeave">
-            <div :class="{ 'active': true, 'modal': true, 'modal-wd': isWide, 'modal-sm': isSmall }" id="modal-id" v-show="show">
+            <div :class="{ 'active': true, 'modal': true, 'modal-wd': isWide, 'modal-sm': isSmall }" id="modal-id" v-show="show" role="dialog" aria-modal="true" aria-labelledby="modal-title">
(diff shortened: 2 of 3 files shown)


psiinon:
Great - many thanks!

psiinon:
Looks good to me :)
Sorry for the delay - was very distracted by ZAPCon.
Could you sign off your commit so that it passes the DCO check and also update the [changelog](https://github.com/zaproxy/zap-hud/blob/main/CHANGELOG.md)

zersiax:
Just amended the final commit with --signoff, that should do it :)

thc202:
Actually all commits need to be signed off.

zersiax:
Not entirely sure how to get that commit to sign off retroactively, I already did what the DCO check suggests, but it seems one commit, a0ab062, did not get signed off on by following that suggestion. what am I missing here?

kingthorin:
You’d have to do an interactive rebase to do each one, in which case you may as well just squash it down to a single signed off commit. (Then you’ll have to force push.)

zersiax:
Yeah, got it. hashtag tryHarder and all that.

thc202 (approved):

thc202:
Thank you!

thc202:
For the record I didn't manually test the change (CI has HUD tests though).

psiinon:
I'm not getting any alerts shown in the HUD now :/
Will double check its not some other local issue...

zersiax:
Yeah I wasn't able to manually test either, so ...very much cowboy coding, this. If this PR turns out to not work well, I might try fixing various things using a tampermonkey script instead

psiinon:
Its ok, local problem :)
Looks like my dev branch is broken for the HUD, the current one fails too. Works fine on 2.10.0.
Also sanity checks with the weekly and thats fine too.

psiinon (approved):

psiinon:
Many thanks!
@zersiax what name + link or twitter acc should we use for you in the credits? 
https://www.zaproxy.org/docs/desktop/credits/

zersiax:
oh, just Florian Beijers (https://twitter.com/zersiax) is fine :)

psiinon:
https://github.com/zaproxy/zap-core-help/pull/353 :)

PR #936, opened 2021-04-26
Use core helper methods to read configs

thc202:
Use the core methods that handle conversion exceptions when reading the
configurations.


kingthorin (approved):
Thanks

psiinon (approved):

PR #955, opened 2021-05-30
chore: Update Gradle to 7.0.2

kingthorin:
Per https://docs.gradle.org/current/userguide/gradle_wrapper.html via
`./gradlew wrapper --gradle-version 7.0.2 --distribution-type all
--gradle-distribution-sha256-sum=13bf8d3cf8eeeb5770d19741a59bde9bd966dd78d17f1bbad787a05ef19d1c2d`

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>


thc202:
It would be better to only merge these PRs until all repos have a PR, to avoid using multiple versions at the same time.

PR #1065, opened 2022-01-20
Improve hud accessibility for screen reader users

zersiax:
Signed-off-by: Florian Beijers <florianbeijers@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Added
+
 - Added accessibility fixes for rapid navigation between panels. 
 - added semantic roles and better labeling for screen reader users
+
+#### Changed
+
 - Bumped ZAP version in build.gradle.kts to v2.11.1
 ## [0.13.0] - 2021-10-06
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-#### Added
-
+### Added
 - Added accessibility fixes for rapid navigation between panels. 
-- added semantic roles and better labeling for screen reader users
-
-#### Changed
-
-- Bumped ZAP version in build.gradle.kts to v2.11.1
+- Added semantic roles and better labelling for screen reader users.
+### Changed
+- Update minimum ZAP version to 2.11.1.
 ## [0.13.0] - 2021-10-06
 ### Changed
 - Update minimum ZAP version to 2.11.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-
+- Added accessibility fixes for rapid navigation between panels. 
+- added semantic roles and better labeling for screen reader users
 ## [0.13.0] - 2021-10-06
 ### Changed
 - Update minimum ZAP version to 2.11.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added accessibility fixes for rapid navigation between panels. 
 - added semantic roles and better labeling for screen reader users
+- Bumped ZAP version in build.gradle.kts to v2.11.1
 ## [0.13.0] - 2021-10-06
 ### Changed
 - Update minimum ZAP version to 2.11.0.
--- a/src/main/zapHomeFiles/hud/display.html
+++ b/src/main/zapHomeFiles/hud/display.html
@@ -33,7 +33,7 @@
         <template id="modal-template">
             <transition name="modal" @after-leave="afterLeave">
             <div :class="{ 'active': true, 'modal': true, 'modal-wd': isWide, 'modal-sm': isSmall }" id="modal-id" v-show="show" role="dialog" aria-modal="true" aria-labelledby="modal-title">
-                <a href="#close" class="modal-overlay" aria-label="Close" @click="close"></a>
+                <a href="#close" class="modal-overlay" aria-label=$t('message.a11y_label_close') @click="close"></a>
                 <div class="modal-container">
                     <div class="modal-header">
                         <slot name="header"></slot>
@@ -56,7 +56,7 @@
         <template id="nav-modal-template">
             <modal :title="title" :show="show" @close="close">
                 <div slot="header">
-                    <button v-if="isBackShowing" class="btn btn-action btn-sm float-left"  @click="back" aria-label=" Back">
+                    <button v-if="isBackShowing" class="btn btn-action btn-sm float-left"  @click="back" aria-label="Back">
                         <i class="icon icon-back"></i>
                     </button>
                 </div>
@@ -277,7 +277,7 @@
 
         <!-- http message modal component -->
         <template id="http-message-modal-template">
-            <nav-modal :title="title" :show="show" size="wide" @close="close" @back="back" :stack="stack" aria-label="HTTP message"> 
+            <nav-modal :title="title" :show="show" size="wide" @close="close" @back="back" :stack="stack" aria-label="HTTP Message"> 
                 <div slot="body">
                     <tabs :activetab="activeTab">
                         <tab :name="$t('message.common_request')" selected="true">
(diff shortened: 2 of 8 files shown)


zersiax:
This PR does a number of things:

- the various iframes are now titled, which makes their function and location clearer. This should help in making the tutorial easier to follow.
- The roles and labels of some of the components have been overwritten for screen reader users only. This will make identification of the various controls easier. Notably, buttons are actually buttons semantically now where they were divs before, and the labels for the various tools are easier to follow now.

Some of the labels are currently English-only, and these changes do nothing for keyboard-only users just yet, but I'm intending to add some more odds and ends as I think of them.

thc202:
The formatting in `inject.js` needs to be fixed: e.g. `571:1 Expected indentation of 2 tabs but found 0.                                  indent`.

thc202:
We can also addressed that, as you prefer.

psiinon:
Thanks for submitting this - we're always very happy to see accessibility improvements.

It should be easy to make the labels I18n, eg change: 
`<modal :title="title" :show="show" @close="close" aria-label="Select Tools">`
to something like
`<modal :title="title" :show="show" @close="close" aria-label="$t('message.label_select_tools')">`
Then add the following line to src/other/resources/UIMessages/UIMessages.properties:
`label_select_tools = Select Tools`

This isnt mandatory for this PR as far as I'm concerned but it seems a shame not to do it at this stage as theres only a few instances right now?

zersiax:
Agreed. Will take some time today to fix the linting errors and add the i18n support. Do we have access to $t() within inject.js for the frame titles, as well?

psiinon:
Possibly not :/ You could try and see - if not we'll have to look into that...

zersiax:
@psiinon  doesn't look like it :( getting $t undefined errors when I try. I will leave those entries alone for now, then, can certainly do the other ones, though. Give me half an hour.

zersiax:
@psiinon  if I add the $t() calls in the aria-label attributes of the elements like yousuggessted, it seems they don't get evaluated properly, or not in time. Example:
			<hud-button icon="<<ZAP_HUD_FILES>>/image/plus.png" name="add-tool" aria-label="$t('message.a11y_label_add_tool')"></hud-button>


This results in the label getting set to that hardcoded string , rather than it getting evaluated. I also tried by doing "{{ $t('message.a11y_label_add_tool'}}"wih similar results.
I also see I didn't seem to have gotten to all the linting errors :) checking

src/main/zapHomeFiles/hud/display.html
```diff
@@ -108,7 +108,7 @@
 
         <!-- select tool modal component -->
         <template id="select-tool-modal-template">
-            <modal :title="title" :show="show" @close="close">
+            <modal :title="title" :show="show" @close="close" aria-label="Select Tools">
```
psiinon, line 111: Worth i18n'ing this and the other ones in this file?

src/main/zapHomeFiles/hud/panel.html
```diff
@@ -15,13 +15,13 @@
 
 	<template id="hud-buttons-template">
 		<div id="hud-buttons" :class="{'d-hide': !isVisible, 'd-visible': isVisible}">
-			<hud-button v-for="tool in tools" :key="tool.name" :label="tool.label" :icon="'<<ZAP_HUD_FILES>>/image/' + tool.icon" :data="tool.data" :name="tool.name"></hud-button>
-			<hud-button icon="<<ZAP_HUD_FILES>>/image/plus.png" name="add-tool"></hud-button>
+			<hud-button v-for="tool in tools" :key="tool.name" :label="tool.label" :icon="'<<ZAP_HUD_FILES>>/image/' + tool.icon" :data="tool.data" :name="tool.name" role="button" :aria-label="label"></hud-button>
+			<hud-button icon="<<ZAP_HUD_FILES>>/image/plus.png" name="add-tool" aria-label="Add Tool"></hud-button>
```
psiinon, line 19: Hopefully these can be i18n'ed as well...

zersiax:
@psiinon  I wrote strings for all of those, but like I stated in the comment above the translations don't appear to be taking :) Could you double-check the syntax you gave me for correctness? If it is correct, we might be running into a limitation of the way ARIA properties on elements are calculated by the browser; JS might be coming in too late to influence the way they are rendered. 
For completeness, if I set the  "add tools" to "$t('message.a11y_label_add_tools')" insead on the aria-label attribute, the screen reader will literally read out $t('message.a11y_label_add_tools) rather than the retrieved value, and I'm not sure why that is happening. I tried variations with adding and removing a : in front of the aria-label property, as well as using {{}} within the double quotes to try to explicitly interpolate the JS, but it doesn't appear to be having an effect in my tests.

psiinon:
Arg, sorry! I'll have a look at that tomorrow:)

psiinon:
Based on this example: https://vuejsdevelopers.com/2019/05/13/accessibility-validation-aria-vuejs/
I've worked out that putting a `:` immediatelly before `aria-label=$t('message.a11y_label_close')` makes it work!
So `:aria-label=$t('message.a11y_label_close')` :)
OK, so I dont really know why thats needed, but it makes it work for me :D

zersiax:
@psiinon confirmed working, I'll make the requisite changes. I also noticed that panel.html wasn't i18n-enabled, I've fixed that.
I'm also noticing zap v2.1.0 is being used by gradle, need me to bump that up to 2.1.1 while I'm here?

psiinon:
2.11.1? :) please!

thc202:
Update the changelog as well :)

src/main/zapHomeFiles/hud/drawer.html
```diff
@@ -140,7 +140,7 @@
                         <drawer-button-settings></drawer-button-settings>
                     </li>
                     <li v-if="tabsVisible" class="tab-item tab-action">
-                        <button class="btn btn-action btn-sm"  @click="toggleOpenClose">
+                        <button class="btn btn-action btn-sm"  @click="toggleOpenClose" aria-label="Open/Close">
```
psiinon, line 143: i18n as well?

src/main/zapHomeFiles/hud/drawer.html
```diff
@@ -154,18 +154,18 @@
...
-        <div :class="{'hud-button': true, 'small': true, 'active': isActive}" @click="$emit('click');" @mouseover="mouseOver" @mouseleave="mouseLeave">
+        <div :class="{'hud-button': true, 'small': true, 'active': isActive}" @click="$emit('click');" @mouseover="mouseOver" @mouseleave="mouseLeave" role="button">
             <img :src="icon" />
             <span>{{label}}</span>
         </div>
     </template>
 
     <template id="drawer-button-settings-template">
-        <drawer-button-template :icon="icon" @click="showHudSettings"></drawer-button-template>
+        <drawer-button-template :icon="icon" @click="showHudSettings" aria-label="Settings"></drawer-button-template>
```
psiinon, line 164: Ditto ;)

src/main/zapHomeFiles/hud/drawer.html
```diff
@@ -154,18 +154,18 @@
...
     </template>
 
     <template id="drawer-button-showhide-template">
-        <drawer-button-template :icon="icon" @click="toggleIsVisible"></drawer-button-template>
+        <drawer-button-template :icon="icon" @click="toggleIsVisible" aria-label="Show/Hide HUD"></drawer-button-template>
```
psiinon, line 168: Ditto ;)

psiinon (commented):
A couple more places that can be internationalised - shame not to do that now?
But otherwise looking good!

zersiax:
good catch, i forgot to put those three back. fixed now.

psiinon (approved):

psiinon:
Many thanks!

psiinon:
@thc202 ok with this PR?

CHANGELOG.md
```diff
@@ -6,7 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-
+- Added accessibility fixes for rapid navigation between panels. 
+- added semantic roles and better labeling for screen reader users
+- Bumped ZAP version in build.gradle.kts to v2.11.1
```
thc202, line 11:
```suggestion
### Added
- Added accessibility fixes for rapid navigation between panels. 
- Added semantic roles and better labelling for screen reader users.

### Changed
- Update minimum ZAP version to 2.11.1.

```

CHANGELOG.md
```diff
@@ -6,7 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [Unreleased]
 
+#### Added
 
+- Added accessibility fixes for rapid navigation between panels. 
+- added semantic roles and better labeling for screen reader users
+
+#### Changed
+
+- Bumped ZAP version in build.gradle.kts to v2.11.1
```
thc202, line 16:
```suggestion
### Added
- Added accessibility fixes for rapid navigation between panels. 
- Added semantic roles and better labelling for screen reader users.
### Changed
- Update minimum ZAP version to 2.11.1.
```

thc202 (approved):

thc202:
Thank you!

PR #1075, opened 2022-01-31
Use server provided by the Network add-on

thc202:
Replace the usage of core `ProxyServer` and related classes with the
`Server` provided by the add-on.

---
The build will fail, the add-on needs to be included in the weekly. The add-on also needs to be released to not require to push the binary to the repo.


thc202:
The build is now passing with the latest weekly.

psiinon (approved):

thc202:
Removed the binary, now using the artifact from Maven Central.

kingthorin (approved):

PR #1102, opened 2022-04-14
fix - npm test error

knassar702:
https://github.com/zaproxy/zap-hud/issues/1103

* using const instead of import
* adding utils.js file to utils.test.js

### DIFF
--- a/src/test/js/utils.test.js
+++ b/src/test/js/utils.test.js
@@ -1,14 +1,764 @@
-import test from 'ava';
+// Public variables
+const test = require("ava");
+const IS_HUD_INITIALIZED = 'isHudInitialized';
+const IS_FIRST_TIME = 'isFirstTime';
+const IS_SERVICEWORKER_REFRESHED = 'isServiceWorkerRefreshed';
 
-// TODO: use clean build system
-// Dynamically load utils functions from utils.js
-// <dirty-hack>
-const fs = require('fs');
+const LOG_OFF = 0;	// Just use for setting the level, nothing will be logged
+const LOG_ERROR = 1;	// Errors that should be addressed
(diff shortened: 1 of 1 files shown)


src/test/js/utils.test.js
```diff
@@ -1,14 +1,764 @@
...
+const LOG_TRACE = 5;	// Very fine grain events, highest level
+const LOG_STRS = ['OFF', 'ERROR', 'WARN', 'INFO', 'DEBUG', 'TRACE'];
 
-const CODE_PATH = './src/main/zapHomeFiles/hud/utils.js';
-const jsCode = fs.readFileSync(CODE_PATH, 'utf-8');
-global.eval(jsCode);
-// </dirty-hack>
+class NoClientIdError extends Error {}
+
+const utils = (function () {
```
thc202, line 17: This does not look right, we should not need to copy the script around to test it, we didn't before (when it worked).

knassar702: I tried to import it by using const or import commands, but it won't work
keep in mind the last commit on this file (before my changes ) imports the file by using `global.eval`, and this doesn't import the utils function correctly, and that's what caused the problem

thc202: I understand that it's broken now but copying does not seem to be the correct fix.

knassar702: Yeah, but could we use this as a temporary solution until we find a better way to import this file?

thc202: IMO it would better to disable the test than merge this workaround (which is no longer testing the actual code but the copied one).

thc202:
Closing, this is not the correct fix, there was no more activity, and there's a new PR (#1296).

PR #1109, opened 2022-05-09
Revert vuejs

psiinon:
Vuejs 3.2 has changed the way it runs and meant that the HUD was broken.
We should update to the latest version but I had a go and just updating the js file names wasnt enough so it will require more work.
They works with Firefox and Chrome on my Mac.

The package-lock.json file was generated using `npm i --package-lock-only`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -211,74 +211,6 @@
         }
       }
     },
-    "@intlify/core-base": {
-      "version": "9.1.10",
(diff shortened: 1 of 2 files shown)


kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #1177, opened 2023-01-16
Added Clear Button to History and WebSockets Tab

### LINKED ISSUE #411: Add option to clear history / websockets tab
This would allow users to ignore older messages they are no longer interested in.

Vitika9:
Fixes #411 
Signed-off-by: Vitika9 <vitika.program@gmail.com>

### DIFF
--- a/src/main/zapHomeFiles/hud/drawer.html
+++ b/src/main/zapHomeFiles/hud/drawer.html
@@ -54,6 +54,11 @@
                         <th class="title hidden">
                             <span>{{ historyItemsFilteredMessage }}</span>
                         </th>
+                        <th class="title">
+                            <span>
+                                <button @click="clearHttpMessages()">{{ clearText }}</button>
+                            </span>
+                        </th>
                     </tr>
                 </thead>
             </table>
@@ -104,6 +109,11 @@
                         </th>
                         <th class="title hidden">
                             <span>{{ websocketsItemsFilteredMessage }}</span>
+                        </th>
+                        <th class="title">
+                            <span>
+                                <button @click="clearWebSocketsMessages()">{{ clearText }}</button>
+                            </span>
                         </th>
                     </tr>
                 </thead>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Added Clear Button to History and WebSockets Tab (Issue 411).
 ### Fixed
 - Prevent exception if no display (Issue 3978).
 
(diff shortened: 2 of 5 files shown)


thc202:
The changelog needs to be updated. I didn't try the changes yet.

psiinon:
I'm having problems getting the HUD to run correctly, which is not a problem with this PR.
Can you post a screenshot of how the button appears?
I'm not seeing it yet, but that could be a local problem...

psiinon:
I can run the HUD via the weekly, so must be a build issue for me.
However when I copy your changes in I dont see the new button anywhere :/
Screenshot 2023-01-18 at 11-33-52 The BodgeIt Store
What am I missing?

Vitika9:
> However when I copy your changes in I dont see the new button anywhere :/ 

The new button is at the top right, I don't know why it isn't visible at your end :(
image

psiinon:
Thanks! I'll have another go...

psiinon:
conflicting now :/

Vitika9:
> conflicting now :/

Done!

Vitika9:
> It seems that clicking the clear button in the History tab also clears the WebSockets tab (and vice versa), is that the expected behaviour? I thought they would work independently (or I'm doing something wrong?). Using `ws.ifelse.io/.ws` for tests.

My bad, I was using the same event for both of them. Fixed now 😀

Vitika9:
@psiinon @thc202 do I need to make more changes? Please review.

src/main/zapHomeFiles/hud/drawer.html
```diff
@@ -54,6 +54,11 @@
                         <th class="title hidden">
                             <span>{{ historyItemsFilteredMessage }}</span>
                         </th>
+                        <th class="title">
+                            <span>
+                                <button @click="clearHttpMessages()">Clear</button>
```
psiinon, line 59: The GUI should not have hard coded display strings in it (like "Clear") - those cannot be i18ned. Instead use something like `:name="$t('message.history_tool')"` - thse tags are defined in src/other/resources/UIMessages/UIMessages.properties

psiinon (changes requested):

src/main/zapHomeFiles/hud/drawer.html
psiinon: The rest of the HTML uses these so you should be able to find plenty of examples, but if not just let us know.
Othert than that the PR seems to work well :)

Vitika9: Done, thanks 😀

psiinon (approved):
LGTM!

thc202 (approved):

thc202:
Thank you!

psiinon:
@Vitika9 FYI we have just released the HUD with this change - I've installed it from the market place and it works well!
Many thanks - really good to see the HUD progressing again.
What are your details for the ZAP credits? 
https://www.zaproxy.org/docs/desktop/credits/#zap-extended-team

Vitika9:
@psiinon I am glad to see my changes reflect there, will contribute more! 😃
Here are my details: Vitika Soni, https://github.com/Vitika9

PR #1311, opened 2023-10-14
Remove iframe scroll (deprecated) 

njmulsqb:
scroll attribute is deprecated on iframe; it is replaced with `overflow: hidden;` CSS property.
Reference: https://support.moonpoint.com/network/web/html/css/iframe-obsolete-elements.php

### DIFF
--- a/src/main/zapHomeFiles/hud/target/inject.js
+++ b/src/main/zapHomeFiles/hud/target/inject.js
@@ -567,7 +567,6 @@ const injection = (function () {
 		const mframe = document.createElement('iframe');
 		mframe.id = MANAGEMENT;
 		mframe.src = ZAP_HUD_FILES + '/file/management.html?url=' + URL + '&frameId=management&tabId=' + tabId;
-		mframe.scrolling = 'no';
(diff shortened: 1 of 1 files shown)


njmulsqb:
Hi,

Kindly take out time to test and merge this small improvement to HUD :)

psiinon:
On my list!

psiinon (approved):
Looks good - I cant tell any change in the UI 😁

thc202:
This makes the bottom panel to have a scrollbar in macOS, while with attribute it didn't have it. (I don't mind much either way, since with the default settings it will not be shown unless scrolling in the frame.)

njmulsqb:
> (I don't mind much either way, since with the default settings it will not be shown unless scrolling in the frame.)

So should we merge this as is?

thc202 (approved):

thc202:
Thank you!

> So should we merge this as is?

That's fine by me but this will be checked further.

psiinon:
I dont get the scrollbar on my mac - maybe its screensize dependant?
In any case, I'm good to merge 😁

